PR #24, opened 2019-11-04
Handle the case of major=0 and minor=0

### LINKED ISSUE #19: TODO: Need to decide what to do if latest release is of the form 0.0.a

DilumAluthge:
Fixes #19 

Suppose that the latest version of a package is `a.b.c`.

If `a` is zero and `b` is zero, then the new compat entry will be `0.0.c`.

Else, the new compat entry will be `a.b`.


DilumAluthge:
Bors merge

PR #25, opened 2019-11-04
Delete CompatHelper.yml

DilumAluthge:
This same file exists somewhere else (https://github.com/bcbi/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml). No need to have duplicate copies.


DilumAluthge:
bors merge

PR #33, opened 2019-11-06
By default, keep the existing compat entries

### LINKED ISSUE #31: CompatHelper should keep lower bound
When updating it should go from
```
[Compat]
Example = "0.1"
```
to
```
[Compat]
Example = "0.1, 0.2"
```
and not as currently
```
[Compat]
Example = "0.2"
```

See https://github.com/mauro3/WhereTheWaterFlows.jl/pull/2#discussion_r342451250 where @fredrikekre pointed this out.

DilumAluthge:
Fixes #31


DilumAluthge:
bors merge

PR #45, opened 2019-11-06
Better documentation when there is no existing compat entry

### LINKED ISSUE #38: It doesn't make sense to say we kept or dropped the existing compat if there was no existing compat

DilumAluthge:
Fixes #38


DilumAluthge:
bors merge

PR #67, opened 2019-11-06
Merge da/develop into master

### LINKED ISSUE #63: Informative commit message
It would be great if the commit message reflected which compat entry was changed, instead of
`Automated commit by CompatHelper.jl` from the branch 
` compathelper/new_version/2019-11-06-06-06-16-381-3275298043`, the commit message would say the same thing as the pull request title.

For instance, when browsing the build log on travis, all I can see is the commit message and the brach name, so it's hard for me to relate to which entry a particular build is testing.

### LINKED ISSUE #66: Cron job will fail with error if one or more dependencies is unregistered
I have added CompatHelper to my package, but it has sent me a lot of emails about it. Where's the problem?
https://github.com/JuliaMusic/MusicXML.jl/runs/291007366

DilumAluthge:
1. PR title and commit message - same content - fixes #63 
2. Don't error if a dependency is unregistered - fixes #66

### DIFF
--- a/src/new_versions.jl
+++ b/src/new_versions.jl
@@ -64,7 +64,7 @@ function make_pr_for_new_version(precommit_hook::Function,
               dep)
     else
         if isnothing(latest_version)
-            @error("The dependency \"name\" was not found in any of the registries", dep)
+            @error("The dependency \"$(name)\" was not found in any of the registries", dep)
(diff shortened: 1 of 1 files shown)


DilumAluthge:
bors merge

PR #96, opened 2019-11-10
Simplify the Travis configuration

DilumAluthge:
Currently, we have two build stages:
1. The first build stage runs only the unit tests.
2. The second build stage runs both the unit tests and the integration tests. This build stage requires a secret Travis environment variable.

Because we are using Bors, both build stages work on all PRs, whether the PR is from within this repo or from a fork.

Therefore, there is really no need to have these two stages. We can just have one build stage that runs both the unit tests and the integration tests.


DilumAluthge:
bors merge

Issue #85, opened 2019-11-09
feature request: local usage


tpapp:
It would be nice to be able to use some features of this package locally, on one's own computer, *on demand*.

Eg I could imagine something like
```julia
julia> import CompatHelper

julia> CompatHelper.show_possible_upgrades("~/code/to/my/repo")
# just list information, eg Foo could be upgraded 0.1.2 => 0.3
# do not modify files

julia> CompatHelper.propose_possible_upgrades("~/code/to/my/repo")
# when there are no uncommitted changes, modify files, otherwise abort
```
The user could then review and test these.

Looking at the sources, I think a lot of the building blocks for this are actually available.

DilumAluthge:
Not a bad idea.

I won't have time to work on this, but if someone else wants to work on it and make a PR, I'd be happy to review the PR.

tpapp:
https://gitlab.com/aplavin/compathelperlocal.jl deals with this effectively, so I would suggest that it is just mentioned in the README.md.

PR #102, opened 2019-11-10
Update manifests by default

christopher-dG:
This finds all Manifest.toml files, activates their environments, and runs `Pkg.update`. 
There is a platform-specific `find` invocation in there, I suppose it should probably go away if we have something like #85. It could be implemented instead with `walkdir` or Glob.jl (although https://github.com/vtjnash/Glob.jl/issues/19 is needed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,24 +32,23 @@ CompatHelper is now installed as a GitHub Action on your repository.
 * Open the specific repository, navigate to the Settings tab, click Actions option, check if the Actions is enabled for this repository.
 
 
-## Updating `Manifest.toml` files after updating compat information
+## Custom pre-commit hooks
 
-If you check in `Manifest.toml` file(s), the PR created by default CompatHelper setup does not run CI with the latest versions of the dependencies.  In this case, it may be a good idea to run `Pkg.update()` via CompatHelper.  This can be done simply using the following snippet instead `run: julia -e 'using CompatHelper; CompatHelper.main()'`:
+CompatHelper supports running a custom function just before commiting changes.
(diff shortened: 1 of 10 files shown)


DilumAluthge:
FYI we should do the package operations in a separate Julia process.

So for example, you'd have something like this:
```julia
function update_manifests()
    list_of_manifest_files = blah blah blah # use walkdir
    for f in list_of_manifest_files
        @info "Updating manifest: $f"
        run(`$(Base.julia_cmd()) -e 'import Pkg; Pkg.activate("$(f)")'; Pkg.update()`)
    end
end
```

DilumAluthge:
Also, we'll need at least some kind of test

christopher-dG:
Can you explain why this needs to happen in a new process?

DilumAluthge:
Here are some snippets from Kristoffer in a Slack conversation in #pkg-usage on Thursday:

> Sounds like you should spawn a new julia process with --project=example/A/ instead. Because if you load stuff in one environment and then change environment and load other things, strange stuff can happen.

> Loading stuff from different environments in one process is just not great.

> Well, you shouldn't really load things from different subprojects in the same process either.

christopher-dG:
Works for me :smile:

DilumAluthge:
Here's a question for you.

If both `path/Project.toml` and `path/Manifest.toml` exist, which is a better idea?
1. `cd` into `path` and run `Pkg.instantiate(); Pkg.update()`
2. `cd` into `path`, delete the `path/Manifest.toml`, and run `Pkg.instantiate(); Pkg.update()`

christopher-dG:
Hm, I'm not sure if there's a difference. But AFAIK, just doing `update` will update the manifest in the right way.

PR #104, opened 2019-11-10
Automatically update all Manifest files

DilumAluthge:
Closes #102 

This is an expanded version of the excellent work by @christopher-dG in #102. This PR adds the following on top of the work in #102:
1. Adds tests.
2. Makes sure that we support custom registries.
3. Run jobs in a "sandbox" that does not have access to most environment variables. That way, if a package has a malicious build step, for example, they cannot steal your GitHub token.

---

I have been thinking about this, and it makes sense that "update all Manifests" should be the default behavior. After all, Dependabot will update your `package.json` and `package-lock.json` files by default. So we should do the same thing. It is easy enough to disable this default behavior by passing a no-op function `() -> ()` as the first positional argument to `CompatHelper.main`.

That being said, because we are changing the default behavior, this is a breaking change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,5 @@
 *.jl.cov
 *.jl.mem
 .DS_Store
-/Manifest.toml
+Manifest.toml
 /dev/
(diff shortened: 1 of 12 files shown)


DilumAluthge:
bors merge=@DilumAluthge,@christopher-dG

DilumAluthge:
bors merge=@DilumAluthge,@christopher-dG

PR #157, opened 2019-11-25
Run CompatHelper once per hour

DilumAluthge:


nalimilan:
What about running only once a day? There's no hurry to update bounds within one hour, anyway it's unlikely that a developer will be there to merge it in less than one day.

DilumAluthge:
> What about running only once a day? There's no hurry to update bounds within one hour, anyway it's unlikely that a developer will be there to merge it in less than one day.

It’s entirely up to the owner of each package. Just edit the `cron` entry to whatever you like :)

I chose one hour as the default because several users told me they wanted a quick turnaround. But I think it’s very reasonable to change it to once a day on your repos.

This cron entry will run CompatHelper once a way. Just edit the `.github/workflows/CompatHelper.yml` file in your repo and change the cron entry to:
```yaml
- cron: '00 00 * * *'
```

DilumAluthge:
See e.g. https://github.com/JuliaData/DataFrames.jl/pull/2044

nalimilan:
Thanks for the PR! I know I can edit the files, but I had no idea of the string to use to run CRON once a day. :-)

I still think once a day would be a better default (if only for ecological reasons), but of course that's up to you.

PR #171, opened 2019-12-29
Simplify CompatHelper.yml example in README.md

tkf:
IIUC, you don't need to use `strategy` as you only need to run CompatHelper with single Julia version (and arch and OS).  This would simplify `CompatHelper.yml` a bit more.

This example seems to work: https://github.com/tkf/BangBang.jl/pull/79


DilumAluthge:
Bors merge

DilumAluthge:
Bors merge

PR #186, opened 2020-03-23
update to Pkg v1.4

johnnychen94:
I think I've mistakenly set the julia version to `"1"` and thus runs the latest 1.4.0 version, which isn't compatible to CompatHelper yet, and get tens of github action error notifications.

An example can be found here: https://github.com/JuliaImages/ImageContrastAdjustment.jl/runs/526166418?check_suite_focus=true

I'm not sure if this fixes all the things, is there a way to run CompatHelper locally? (Looks like not)

### DIFF
--- a/src/stdlib.jl
+++ b/src/stdlib.jl
@@ -1,7 +1,15 @@
 import UUIDs
 
+if VERSION < v"1.4"
+    stdlibs = Pkg.Types.stdlib
+else
+    # renamed to stdlibs
+    # https://github.com/JuliaLang/Pkg.jl/pull/1559
+    stdlibs = Pkg.Types.stdlibs
+end
+
 # This function is based off of a similar function here:
 # https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/src/RegistryCI.jl
 function gather_stdlib_uuids()
-    return Set{UUIDs.UUID}(x for x in keys(Pkg.Types.stdlib()))
+    return Set{UUIDs.UUID}(x for x in keys(stdlibs()))
 end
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Can you bump the version number? Because this is a bug fix, You can just bump the patch number.

DilumAluthge:
bors merge

src/stdlib.jl
```diff
@@ -1,7 +1,15 @@
 import UUIDs
 
+if VERSION < v"1.4"
+    stdlibs = Pkg.Types.stdlib()
```
fredrikekre:
```suggestion
    stdlibs = Pkg.Types.stdlib
```

src/stdlib.jl
```diff
@@ -1,7 +1,15 @@
...
+else
+    # renamed to stdlibs
+    # https://github.com/JuliaLang/Pkg.jl/pull/1559
+    stdlibs = Pkg.Types.stdlibs()
```
fredrikekre:
```suggestion
    stdlibs = Pkg.Types.stdlibs
```

src/stdlib.jl
```diff
@@ -1,7 +1,15 @@
...
+end
+
 # This function is based off of a similar function here:
 # https://github.com/JuliaRegistries/RegistryCI.jl/blob/master/src/RegistryCI.jl
 function gather_stdlib_uuids()
-    return Set{UUIDs.UUID}(x for x in keys(Pkg.Types.stdlib()))
+    return Set{UUIDs.UUID}(x for x in keys(stdlibs()))
```
fredrikekre: You need the two changes above, or the following:
```suggestion
    return Set{UUIDs.UUID}(x for x in keys(stdlibs))
```

johnnychen94: Ooops my mistake.

DilumAluthge:
Bors merge

DilumAluthge:
Thank you @johnnychen94 and @fredrikekre

Issue #212, opened 2020-06-02
README example out of date?


ericphanson:
I've been using the example 3
```julia
run: julia -e 'using CompatHelper; CompatHelper.main( (; registries) -> CompatHelper._update_manifests(pwd(); registries = registries) )'
```
so CompatHelper does not update a particular Manifest (which I want to keep for reproducibility), but it's been failing: https://github.com/ericphanson/GuessworkQuantumSideInfo.jl/runs/729164733?check_suite_focus=true. I'm not quite sure how to change it to fix the issue.

DilumAluthge:
Try replacing `pwd()` with `String[pwd()]`.

If that fixes the issue, then we can update the README

ericphanson:
Ok, I'll give it a try!

ericphanson:
Progress! https://github.com/ericphanson/GuessworkQuantumSideInfo.jl/runs/732883246?check_suite_focus=true#step:3:27

DilumAluthge:
Set the `delete_old_manifests` keyword argument to the `_update_manifests` function.

ericphanson:
Would you mind explaining what that does? I don't really get how it could update the manifests but not delete them? Or does it the delete the ones it doesn't update?

I think for me the better solution is actually to just add a github actions CI job to test the project with that manifest separately and then allow CompatHelper to make PRs to update it.

ericphanson:
Ah okay, that makes a lot of sense! I'll send a PR over to update the example and add that info.

PR #214, opened 2020-06-03
Update example

ericphanson:
Closes https://github.com/bcbi/CompatHelper.jl/issues/212

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
-version = "1.9.0"
+version = "1.9.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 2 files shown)


DilumAluthge:
bors merge

ericphanson:
Hold on, I had a better idea-- what about using the `path-ignore` so Version Vigilante doesn't care about stuff that tags shouldn't matter for? I pushed a commit to show what that would be

DilumAluthge:
Can you explain the latest changes?

DilumAluthge:
> Hold on, I had a better idea-- what about using the `path-ignore` so Version Vigilante doesn't care about stuff that tags shouldn't matter for? I pushed a commit to show what that would be

I see.

Unfortunately, Bors is not that smart, so if VersionVigilante doesn't report a status, Bors will never merge a PR.

ericphanson:
Let me know if you want me to revert that last commit, but the idea is that you shouldn't need patch bumps for readme's because it never gets into the code, right? So version vigilante can just check changes to the code/docs files which are the ones that need a version bump.

ericphanson:
Ah ok, I'll revert it then

DilumAluthge:
It would be nice to figure out how to get this to work with Bors, but yeah for now it won't work.

ericphanson:
I'm glad Bors was smart enough to cancel when I pushed after you said to merge :)

DilumAluthge:
bors merge

PR #221, opened 2020-07-01
GitHub Enterprise Suport

tk3369:
This PR adds a new host parameter to the main function. When it is specified, it triggers a different logic to authenticate with GitHub. Because the API to call GitHub requires a new leading argument, I have wrapped the several GitHub functions as well. It has been tested in my private environment.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,12 +1,13 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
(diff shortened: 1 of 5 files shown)


DilumAluthge:
I don't like the use of global variables here.

We should create an `API` object at the beginning, and then just explicitly pass that around to every method.

If the user specifies an `API` object, we use that. If they don't, we provide the default `API` object for `GitHub.com`.

tk3369:
Agree that it's a quick and dirty solution at the moment.  I will come back to this when I find more time.

racinmat:
I hope this PR is doing the thing you're discussing here: https://github.com/bcbi/CompatHelper.jl/pull/227

PR #241, opened 2020-08-18
Add action.yml

SaschaMann:
(still testing, no need for review)


DilumAluthge:
Since CompatHelper is nearing end of life, I'm going to close this in favor of the impending Dependabot revolution.

PR #242, opened 2020-08-19
Usage of package server to download registry for Julia 1.4 and newer.

### LINKED ISSUE #209: CompatHelper doesn't respect package server
CompatHelper tries to clone registry even when `JULIA_PKG_SERVER` is set, specifically in
https://github.com/bcbi/CompatHelper.jl/blob/master/src/get_latest_version_from_registries.jl#L17 which leads to completely ignoring the set `JULIA_PKG_SERVER` variable.

racinmat:
Fixes #209 #244 
It's a bit mess but I couldn't figure out better way and it's working.
I'm able to run CompatHelper behind corporate firewall with my own registry on GitHub Enterprise with this change.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
-version = "1.13.5"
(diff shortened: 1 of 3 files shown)


racinmat:
@DilumAluthge is it ok like this or should it be split into multiple smaller PRs? (technically this is fixing 2 different issues)

DilumAluthge:
Yeah can you split it up into two different PRs, one for each issue that you are fixing?

PR #246, opened 2020-08-21
using package server to download registry

### LINKED ISSUE #209: CompatHelper doesn't respect package server

racinmat:
Fixes #209 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
-version = "1.13.7"
+version = "1.14.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
-version = "1.13.6"
+version = "1.13.7"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/bors.toml
+++ b/bors.toml
@@ -3,8 +3,8 @@ block_labels = [ "DO NOT MERGE" ]
 delete_merged_branches = true
 
 status = [
-    "codecov/patch",
-    "codecov/project",
+    # "codecov/patch",
+    # "codecov/project",
     "Travis CI - Branch",
     "VersionVigilante",
 ]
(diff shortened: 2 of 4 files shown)


racinmat:
@DilumAluthge  better now?

DilumAluthge:
Thanks!

I'll do a more thorough review this week or next week.

DilumAluthge:
bors try

racinmat:
Sorry, typo, fixed.

DilumAluthge:
bors try

racinmat:
Sorry, another one. No more books.

DilumAluthge:
bors try

racinmat:
fixed

DilumAluthge:
bors try

racinmat:
I had probably some bad day, now it should pass.

DilumAluthge:
bors try

racinmat:
Any idea how to test this?

DilumAluthge:
We can leave it untested for now.

I'll do one last review this week and then I'll merge.

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge (approved):

DilumAluthge:
bors merge

DilumAluthge:
@racinmat Thanks for your patience!

DilumAluthge:
bors merge

fredrikekre:
> The reason I want to keep the default behavior as cloning over git is this: often, the pkg server does not serve the latest registry. For example, recently, it was serving a registry that was more than 2 hours old. So cloning over git is still the most reliable way to get the registry.

IMO that is an argument for having this default, now the PR that CompatHelper submits just fails instead, since the PkgServer is the default for CI.

racinmat:
Wait, but PkgServer is not default for CI, this PR made in only possible to use it, as in opt-in if you add kwarg to `main` function. 
The default behavior has been left unchanged.

KristofferC:
> Wait, but PkgServer is not default for CI, this PR made in only possible to use it, as in opt-in if you add kwarg to main function.

It is on Julia 1.5+

racinmat:
Maybe we are talking about different things. Yes, Julia Pkg uses by default PkgServer on version 1.5+, but in CompatHelper there is explicit cloning: https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/src/get_latest_version_from_registries.jl#L6
regardless of version.

fredrikekre:
> CompatHelper there is explicit cloning

Yes, that is the problem -- I think this functionality that you implemented should be default.

DilumAluthge:
> now the PR that CompatHelper submits just fails instead, since the PkgServer is the default for CI.

That's a good point

DilumAluthge:
#252

KristofferC:
There's also the problem that there are many pkg servers and they often don't serve the exact same copy of the registry.

PR #251, opened 2020-09-02
Add a flag to not bump compat entries containing an equality specifier

### LINKED ISSUE #250: Don't make a PR to packages that use an equality specifier
I have a request based on a PR I just closed this morning:
> This pull request changes the compat entry for the `FFTW` package from `=1.2.2` to `=1.2.2, 1.2`.

We have an equality specifier on `FFTW` because there are bugs in later versions that prevent us from using them; we must use precisely `FFTW = "=1.2.2"`. It doesn't make sense to me why CompatHelper should change the compat entry in this case, since I have explicitly declared that I want one exact version of the package. My request is that CompatHelper does not make a PR changing the compat entry for this package because I used an equality specifier. Does this make sense to anyone else? Am I missing a use case that this was designed for?

Octogonapus:
This PR adds a kwarg `bump_compat_containing_equality_specifier` that can be set to `false` to avoid bumping compat entries containing an equality specifier.

I didn't see any unit tests that test bumping compat entries; let me know how I can test this.

Closes #250.


src/new_versions.jl
```diff
@@ -79,6 +82,18 @@ function make_pr_for_new_version(api::GitHub.GitHubAPI,
...
+    elseif !bump_compat_containing_equality_specifier &&
+            !isnothing(current_compat_entry_verbatim) &&
+            contains(current_compat_entry_verbatim, '=') &&
+            !contains(current_compat_entry_verbatim, '>') &&
+            !contains(current_compat_entry_verbatim, '<')
+        # To check for an equality specifier (but not an inequality specifier) we look for an equals sign without any
+        # symbols used for an inequality specifier that would also include an equals sign. Namely, greater than and
+        # less than. Other specifiers containing symbols like ≥ shouldn't parse as containing an equals sign.
+        @info("skipping compat entry because it contains an equality specifier",
+            current_compat_entry_verbatim,
```
Octogonapus, line 94: @DilumAluthge I'm not sure if `current_compat_entry_verbatim` will contain the entire compat entry line (e.g., `FFTW = "=1.2.2"`). If that is the case, this needs to be changed.

src/new_versions.jl
```diff
@@ -79,6 +82,18 @@ function make_pr_for_new_version(api::GitHub.GitHubAPI,
               latest_version,
               name,
               dep)
+    elseif !bump_compat_containing_equality_specifier &&
+            !isnothing(current_compat_entry_verbatim) &&
+            contains(current_compat_entry_verbatim, '=') &&
+            !contains(current_compat_entry_verbatim, '>') &&
+            !contains(current_compat_entry_verbatim, '<')
```
Octogonapus, line 89: @DilumAluthge I'm not sure if `current_compat_entry_verbatim` will contain the entire compat entry line (e.g., `FFTW = "=1.2.2"`). If that is the case, this needs to be changed.

DilumAluthge:
Thanks! I need to finish reviewing another PR, and then I'll work on this one.

src/new_versions.jl
DilumAluthge: `current_compat_entry_verbatim` would only be e.g. `"=1.2.2"`. So this should be fine.

DilumAluthge:
bors try

DilumAluthge:
bors try

DilumAluthge:
When I have some time, I'll do one final review of this, and then I'll merge it.

DilumAluthge:
bors merge

PR #262, opened 2020-12-02
Suggest in automatic PR to run tests with --depwarn=yes

mauro3:
Inspired by https://discourse.julialang.org/t/why-is-depwarn-no-by-default/51094


fredrikekre:
This is already default when running `Pkg.test` on 1.5(?) and higher.

mauro3:
Ok, then maybe suggest to look at the tests to see whether there are warnings.  Or is the setting `--depwarn=error`?

DilumAluthge:
Is there a particular reason why CompatHelper should be making these recommendations? It seems like this could be a recommendation in general. After all, a package can add a deprecation warning when going from 0.1.0 to 0.1.1.

PR #265, opened 2021-02-01
only try to remove tmp_dir if it's still there

### LINKED ISSUE #264: with_tmp_dir atexit hook fails
Hi,

I just noticed that the `atexit` hook fails for `with_tmp_dir` because the directory is already gone.
I think the problem is `walkdir` in `prepare_for_deletion` which throws an error by default if it can't `readdir` the directory.
Maybe providing a custom error handler to `walkdir` could fix this?

You can see the error e.g. in [this CompatHelper run](https://github.com/fkastner/NiceNumbers.jl/runs/1801953405?check_suite_focus=true).

fkastner:
This should fix #264.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge <dilum@aluthge.com>", "Brown Center for Biomedical Informatics"]
-version = "1.16.4"
+version = "1.16.5"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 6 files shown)


KristofferC:
`mktempdir` should do this by itself since 1.3 I think.

fkastner:
So `with_tmp_dir` should probably be removed altogether and changed to `mktempdir`?

KristofferC:
I would say so

fkastner:
`mktempdir` has the `cleanup` keyword only since 1.3 but this package still supports 1.2.
Should support for 1.2 be dropped then? What would be the alternative strategy?

KristofferC:
Personally, I don't really see a need to support 1.2 since this is just a bot that runs.

fkastner:
I removed all calls to `rm` where the file/folder is created with `mktemp` or `mktempdir`. In particular the `atexit` hooks are gone now, too. In one case I changed from manually creating a file to using `mktemp` to avoid having to cleanup manually afterwards.

Also the `with_temp_dir` and also `prepare_for_deletion` were removed since they are no longer needed now.

DilumAluthge:
I think #280 should fix all of these issues. Let me know if you still notice any issues after #280 is merged and a new release is registered.

PR #269, opened 2021-03-16
Support Julia 1.7

timholy:
RegistrySpec moved from Pkg.Types to Pkg.Registry.
Since Pkg.RegistrySpec works on all Julia versions back to
1.2 (the earliest version supported), we don't need explicit
version checks if we use this more general way of accessing
the type.


DilumAluthge (approved):

DilumAluthge:
CI is currently on Travis, and it will take a while for me to migrate to GHA because of the complexity of the integration test setup. So I'll YOLO merge thus.

DilumAluthge:
Can you bump the version number?

DilumAluthge (approved):

PR #279, opened 2021-04-04
Recommend that people use their existing `DOCUMENTER_KEY`

### LINKED ISSUE #233: Should we just recommend that people use their existing DOCUMENTER_KEY?
Should we just have that as the default? Similar to what Tim suggested in https://github.com/JuliaRegistries/TagBot/pull/162?

cc: @timholy
cc: @christopher-dG

DilumAluthge:
Fixes #233


DilumAluthge:
bors merge

PR #292, opened 2021-04-21
If a private key is provided, always check it (even if we don't open any new PRs)

DilumAluthge:
Before this PR, if a private key was provided, we would only check it (e.g. try to Base64-decode it, etc.) if we needed to open a new PR.

This led to confusing behavior where some CompatHelper jobs would succeed and others would fail.

After this PR, If a private key is provided, we will always check it (even if we don't open any new PRs). Therefore, if there is a problem with your private key, all CompatHelper jobs will fail.

This should make debugging slightly easier.

cc: @ablaom


DilumAluthge:
bors merge

DilumAluthge:
bors merge

DilumAluthge:
bors merge

PR #296, opened 2021-05-07
NFC: Clean up code formatting and styling

mattBrzezinski:
## Overview of Changes
- Replaced usages of `import` with `using`
- Remove whitespacing around kwargs and their default values
- Formatted function parameters
- Added whitespacing around large blocks of code to increase readability
- Removed variable assignment before `return` statement
- Added `nothing` to empty `catch` statements
- Converted small functions into one liners when applicable
- The only refactor change is in `src/envdict.jl` for the `_generate_env_dict()` function.

A follow up PR around the `test` directory will come shortly.

### DIFF
--- a/src/CompatHelper.jl
+++ b/src/CompatHelper.jl
@@ -3,6 +3,7 @@ module CompatHelper
 using Base64
 using Dates
 using GitHub
+using HTTP
 using Pkg
 using Printf
 using TOML


src/envdict.jl
```diff
@@ -1,16 +1,16 @@
...
-                            JULIA_DEPOT_PATH)
+function _generate_env_dict(
+    original_env::AbstractDict;
+    JULIA_DEPOT_PATH
+)
     result = Dict{String, String}()
-    if haskey(original_env, "HTTP_PROXY")
-        result["HTTP_PROXY"] = deepcopy(original_env["HTTP_PROXY"])
+
+    for var in ["HTTP_PROXY", "HTTPS_PROXY", "JULIA_PKG_SERVER", "PATH"]
```
fchorney, line 7: you put `PATH` here but left out `JULIA_DEPOT_PATH`, but both seem to deepcopy without the check in the original code.

fchorney: oh yeah I see I didn't notice that. Very good

fchorney (approved):

DilumAluthge:
bors try

DilumAluthge:
Does this bring us into compliance with Blue Style? If so, it would be great to add the Blue Style badge to the README.

DilumAluthge:
Looks like you need to import HTTP?

DilumAluthge:
Also, can you bump the version number in the `Project.toml` file?

If you are planning on making multiple non-breaking PRs in a short period of time, you could e.g. bump to `2.1.0-DEV`. Then, once you are done with PRs, you bump to `2.1.0` and make a new release.

Analogously, if you are planning on making multiple breaking PRs in a short period of time, you could e.g. bump to `3.0.0-DEV`. Then, once you are done with PRs, you bump to `3.0.0` and make a new release.

src/main.jl
```diff
@@ -1,19 +1,19 @@
-const default_registries = Pkg.RegistrySpec[Pkg.RegistrySpec(name = "General",
-                                                             uuid = "23338594-aafe-5451-b93e-139f81909106",
-                                                             url = "https://github.com/JuliaRegistries/General.git")]
+const DEFAULT_REGISTRIES = Pkg.RegistrySpec[Pkg.RegistrySpec(name = "General", uuid = "23338594-aafe-5451-b93e-139f81909106", url = "https://github.com/JuliaRegistries/General.git")]
```
GunnarFarneback, line 1: I'm just not buying that this is a style improvement. But I have no involvement in this code so consider it a drive-by comment.

src/CompatHelper.jl
```diff
@@ -1,14 +1,14 @@
...
-import UUIDs
+using Base64
+using Dates
+using GitHub
+using Pkg
+using Printf
+using TOML
+using TimeZones
+using UUIDs
+
```
GunnarFarneback, line 11: I guess it's fine with these packages and the function names used in this code but in general it's not NFC to change `import` to `using`.

PR #302, opened 2021-05-10
Run the integration tests on both Julia v1 and Julia nightly

DilumAluthge:
The GitHub Actions virtual environments always contain the latest stable release of Julia.

Whenever a new stable version of Julia is released, the GitHub Actions staff usually update the virtual environments relatively quickly.

Therefore, we need to make sure that CompatHelper doesn't start breaking whenever the virtual environments are updated to a new version of Julia.

One step we can take is to run the integration tests on both Julia v1 and Julia nightly.

### DIFF
--- a/.github/workflows/ci_integration.yml
+++ b/.github/workflows/ci_integration.yml
@@ -26,6 +26,7 @@ jobs:
           - ubuntu-latest
         version:
           - '1'
+          - 'nightly'
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


DilumAluthge:
bors merge

DilumAluthge:
bors merge-

DilumAluthge:
This will bring the total CI time per PR to approx 30 minutes, which is too long.

PR #305, opened 2021-05-14
Implement fetching PullRequests

### LINKED ISSUE #304: Restore test that the major version in the workflow file matches the major version in the Project.toml file
The documentation home page shows the workflow file at `.github/workflows/CompatHelper.yml`. Since this file is shown to users as part of the installation instructions, the major version number (`version = "..."`) should always be the same as the major version number in `Project.toml`.

(We also check that the major version is greater than or equal to 1, since it is not good to do e.g. `version = "0"`).

@mattBrzezinski Can you add that test back to the test suite? It used to be here: 

https://github.com/JuliaRegistries/CompatHelper.jl/blob/887478129edb0b157f4b72899b3f5fcf704a6906/test/unit-tests.jl#L1-L11

mattBrzezinski:
Fixes #304 

## Overview
This merge request is targeting **v3-dev**. This is a first implementation for gathering pull requests, there are a few things which need to be re-worked here, but it requires re-working the `GitForge.@pagination` functionality.

Things that we need to change and add in here:
- Ability to mock requests, currently the tests run off the basis the both the Julia and GitLab projects have merge requests in them.
- The tests do not have authenticated requests, so we can hit limitations here. I'm not sure of the best bet going forward here should be. Since we at Invenia are working on this project and taking ownership, we can create an account and some PATs as part of our efforts to keep GH Actions alive. See more on that [here](https://github.com/invenia/KeepActionsAlive).
- The pagination functionality with GitLab returns a duplicate of every PR, as a work around for now I've introduced functionality to compare `GitLab.MergeRequest` objects as it does not work currently. See https://github.com/JuliaWeb/GitForge.jl/issues/24

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,11 +9,13 @@ Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 
 [compat]
 GitForge = "0.1"
+Mocking = "0.7"
 julia = "1.6"
 
 [extras]
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "TOML"]
+test = ["Aqua", "Test", "TOML"]
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ julia = "1.6"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 
 [targets]
-test = ["Test"]
+test = ["Test", "TOML"]
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,7 @@ GitForge = "8f6bce27-0656-5410-875b-07a5572985df"
 Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 
 [compat]
+Aqua = "0.5"
 GitForge = "0.1"
 Mocking = "0.7"
 julia = "1.6"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,12 @@ uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge", "Brown Center for Biomedical Informatics", "contributors"]
 version = "3.0.0"
 
+[deps]
+GitForge = "8f6bce27-0656-5410-875b-07a5572985df"
+Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
+
 [compat]
+GitForge = "0.1"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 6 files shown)


DilumAluthge:
>     * The tests do not have authenticated requests, so we can hit limitations here. I'm not sure of the best bet going forward here should be.

This is how we currently have it set up for the CompatHelper integration tests:
1. I created a test user: https://github.com/bcbi-test-user
2. I created a test repo: https://github.com/bcbi-test/compathelper_integration_test_repo.jl
3. A personal access token for the `bcbi-test-user` is available inside the integration tests via `TEST_USER_GITHUB_TOKEN = ENV["BCBI_TEST_USER_GITHUB_TOKEN"]`
4. During the integration tests, we:
    - Create branches
    - Create pull requests
    - Run the integration tests
    - Delete the branches (which automatically closes the pull requests)

This approach has worked well, and I would recommend that we keep using it.

For GitHub, I think we can continue to use the `ENV["BCBI_TEST_USER_GITHUB_TOKEN"]` PAT. For GitLab, if you guys want to create a test user and a test repo, and send me a GitLab PAT via encrypted email, I can add the PAT to this repo. Just send me the PAT as well as the name you want to use for the secret environment variable.

DilumAluthge:
bors try

DilumAluthge:
We'll need a `[compat]` entry for Mocking.

Might be good to use something like Aqua (https://github.com/JuliaRegistries/CompatHelper.jl/issues/307) to make sure we don't miss stuff like that.

mattBrzezinski:
Added a fix for #304 in `6d6f5c7`

mattBrzezinski:
Added a fix for #307 in `65de4d4`

DilumAluthge:
We should probably also have `[compat]` entries for test dependencies.

DilumAluthge:
bors try

mattBrzezinski:
> We should probably also have `[compat]` entries for test dependencies.

Not a bad idea, I'll add those in tomorrow. Was following out conventions at Invenia, although now that I think about it, not too sure why we're not doing this ourselves. 🤔

DilumAluthge:
> > We should probably also have `[compat]` entries for test dependencies.
> 
> Not a bad idea, I'll add those in tomorrow. Was following out conventions at Invenia, although now that I think about it, not too sure why we're not doing this ourselves. 🤔

Most people don't have `[compat]` entries for test dependencies. The two reasons are probably:
1. AutoMerge does not require it
2. CompatHelper currently does not make PRs to bump `[compat]` entries for test-only dependencies. (Although we probably should.)

But I think it's important. Otherwise, if a test-only dependency makes a breaking release, it might break your test suite.

mattBrzezinski:
> > > We should probably also have `[compat]` entries for test dependencies.
> > 
> > 
> > Not a bad idea, I'll add those in tomorrow. Was following out conventions at Invenia, although now that I think about it, not too sure why we're not doing this ourselves. 🤔
> 
> Most people don't have `[compat]` entries for test dependencies. The two reasons are probably:
> 
>     1. AutoMerge does not require it
> 
>     2. CompatHelper currently does not make PRs to bump `[compat]` entries for test-only dependencies. (Although we probably should.)
> 
> 
> But I think it's important. Otherwise, if a test-only dependency makes a breaking release, it might break your test suite.

Yeah that's a good point! Bringing this up elsewhere. Just added in an entry for `Aqua` in `ed219a0`.

mattBrzezinski:
bors try

mattBrzezinski:
Any other thoughts on this? @DilumAluthge @fchorney
Or good to go?

DilumAluthge:
bors merge

PR #310, opened 2021-05-20
Added PullRequest and utility functions

mattBrzezinski:
Began adding in the pull request and some utility functionalities. One thing to note here is there are no tests for the `set_git_identity` function, I'm a bit unsure of how to test this, I'll spend some time thinking about it though.


mattBrzezinski:
bors try

fchorney:
To test `set_git_identity` you could potentially just run git queries after to make sure it was set correctly. 
without the last value, `git config user.name` will return the value it has stored. 

`run(Cmd(String["git", "config", "user.name"]))`
I forget how to pull the results of a run into a string but I think that should let you test it.

src/utilities/utilities.jl
```diff
@@ -0,0 +1,2 @@
+private_compat_helper(; env::AbstractDict=ENV) = haskey(env, "PRIVATE_COMPAT_HELPER") && env["PRIVATE_COMPAT_HELPER"] != "false"
```
fchorney, line 1: These env queries use the string `PRIVATE_COMPAT_HELPER` but in `CompatHelper.jl` you create a const for `PRIVATE_COMPAT_HELPER` that points to `"COMPATHELPER_PRIV"` should this be using the const instead of the string here?

mattBrzezinski: Yes, must have missed this going over it in the last few days. Revising now.

mattBrzezinski:
> To test `set_git_identity` you could potentially just run git queries after to make sure it was set correctly.
> without the last value, `git config user.name` will return the value it has stored.
> 
> `run(Cmd(String["git", "config", "user.name"]))`
> I forget how to pull the results of a run into a string but I think that should let you test it.

Added the test in `402b7e0`

test/utilities/ci.jl
```diff
@@ -0,0 +1,47 @@
...
+
+@testset "set_git_identity" begin
+    function read_git_config()
+        try
+            name = strip(read(`git config user.name`, String))
+            email = strip(read(`git config user.email`, String))
+
+            return name, email
+        catch
+            return "", ""
```
fchorney, line 30: maybe put a note here saying that this might happen if the commands return nothing
Might be worth putting the user.name and user.email requests in different try catches as well? Not too sure when one might work but the other wont, but it could be possible.

DilumAluthge:
I've been thinking about the `set_git_identity` function.

The problem with `set_git_identity` is that it causes global effects that persist. This is fine for GitHub-hosted runners. But if you are e.g. using a self-hosted runner with an configuration that persists between CI jobs, then the effects of `set_git_identity` will be seen in subsequent CI jobs.

So I think we should just get rid of the `set_git_identity` function. And we should not e.g. run `git config --global user.name` or `git config --global user.email`.

Instead, we should just pass the option each time we run a Git command, e.g. `git -c "user.name=Your Name" -c "user.email=Your email" commit`. Then there are no global side effects.

mattBrzezinski:
> I've been thinking about the `set_git_identity` function.
> 
> The problem with `set_git_identity` is that it causes global effects that persist. This is fine for GitHub-hosted runners. But if you are e.g. using a self-hosted runner with an configuration that persists between CI jobs, then the effects of `set_git_identity` will be seen in subsequent CI jobs.
> 
> So I think we should just get rid of the `set_git_identity` function. And we should not e.g. run `git config --global user.name` or `git config --global user.email`.
> 
> Instead, we should just pass the option each time we run a Git command, e.g. `git -c "user.name=Your Name" -c "user.email=Your email" commit`. Then there are no global side effects.

This makes sense, I'll revise the code in the next commit to reflect this changes.

mattBrzezinski:
bors try

DilumAluthge:
I don't think that `private_compat_helper`, `PRIVATE_COMPAT_HELPER`, etc. are the most descriptive names here.

A better name might be `has_ssh_private_key` or `has_ssh_deploy_key` or something like that.

fchorney (approved):

mattBrzezinski:
I'll address these failing CC stuff tomorrow.

PR #311, opened 2021-05-26
Project Dependencies

mattBrzezinski:
Reintroducing the [get_project_dependencies.jl](https://github.com/JuliaRegistries/CompatHelper.jl/blob/a0581339bab9a49b8da58208e30dc771b9fdbb52/src/get_project_deps.jl) code back into the `v3-dev` branch.

After reading through the code it does not seem necessary to return back:

> dep_to_current_compat_entry, dep_to_current_compat_entry_verbatim, dep_to_latest_version, deps_with_missing_compat_entry

Instead I've just simplified it down to return a `Set{CompatEntry}`. The `deps_with_missing_compat_entry` was never actually used it seems? It was retrieved, passed around and when it got to [new_versions.jl](https://github.com/JuliaRegistries/CompatHelper.jl/blob/a0581339bab9a49b8da58208e30dc771b9fdbb52/src/new_versions.jl#L7) it wasn't used.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,9 @@ version = "3.0.0"
 [deps]
 GitForge = "8f6bce27-0656-5410-875b-07a5572985df"
 Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Aqua = "0.5"
--- a/src/CompatHelper.jl
+++ b/src/CompatHelper.jl
@@ -3,7 +3,11 @@ module CompatHelper
 using GitForge
 using GitForge: GitHub, GitLab
 using Mocking
+using Pkg
+using TOML
+using UUIDs
 
+include("dependencies.jl")
 include("pull_requests.jl")
 
 end # module
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -1,0 +1,101 @@
+struct Package
+    name::String
+    uuid::UUIDs.UUID
+end
+
+mutable struct CompatEntry
+    package::Package
+    version_spec::Union{Pkg.Types.VersionSpec, Nothing}
+    version_verbatim::Union{String, Nothing}
+
+    function CompatEntry(p::Package)
+        return new(p, nothing, nothing)
+    end
+end
+
+const LOCAL_REPO_NAME = "REPO"
+
+gather_stdlib_uuids() = Set{UUIDs.UUID}(x for x in keys(Pkg.Types.stdlibs()))
+git_clone(url::AbstractString, local_path::AbstractString) = run(`git clone $(url) $(local_path)`)
+git_checkout(branch::AbstractString) = run(`git checkout $(branch)`)
+
+
+"""
+    add_compat_section(project::AbstractDict)
+
+Add a k/v "compat"=>Dict{Any, Any}() if one does not exist.
+"""
+function add_compat_section!(project::AbstractDict)
+    if !haskey(project, "compat")
+        project["compat"] = Dict{Any, Any}()
+    end
+
+    return project
+end
+
+
+"""
+    get_project_deps(
+        api::GitHub.GitHubAPI, clone_hostname::AbstractString, repo::GitHub.Repo;
+        subdir::AbstractString="", include_jll::Bool=false
+    )
+
+Get a `Set{CompatEntry}` for all dependencies of `repo`.
+"""
+function get_project_deps(
+    api::GitHub.GitHubAPI, clone_hostname::AbstractString, repo::GitHub.Repo;
+    subdir::AbstractString="", include_jll::Bool=false
+)
+    mktempdir() do f
+        url_with_auth = "https://x-access-token:$(api.token)@$(clone_hostname)/$(repo.full_name).git"
+        local_path = joinpath(f, LOCAL_REPO_NAME)
+        @mock git_clone(url_with_auth, local_path)
+
+        # Get all the compat dependencies from the local Project.toml file
+        project_file = @mock joinpath(local_path, subdir, "Project.toml")
+        deps = get_project_deps(project_file; include_jll=include_jll)
+
+        return deps
+    end
+end
+
+
+"""
+    get_project_deps(project_file::AbstractString; include_jll::Bool=false)
+
+Get a Set{CompatEntry} for all dependencies in the Compat section of the project_file.
+Exclude any STDLIB packages and JLL dependencies (unless specified).
+"""
+function get_project_deps(project_file::AbstractString; include_jll::Bool=false)
+    project_deps = Set{CompatEntry}()
+    project = TOML.parsefile(project_file)
+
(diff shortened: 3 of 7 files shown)


mattBrzezinski:
bors try

DilumAluthge:
We'll need to pass e.g. `-c "user.name=Your Name" -c "user.email=Your email"` to every Git command. This allows us to provide the user name and user email to Git without any permanent side effects.

mattBrzezinski:
bors try

mattBrzezinski:
bors try

fchorney:
> We'll need to pass e.g. `-c "user.name=Your Name" -c "user.email=Your email"` to every Git command. This allows us to provide the user name and user email to Git without any permanent side effects.

What do we think would be the best way to save "Your Name" and "Your email" to use it in multiple places?  Would we make the user set a few ENVVARS with those values? Maybe have the user put `set_git_config(name, email)` and then save that somewhere to use later? 

Potentially it might be nice to make some sort of command to call git so you don't need to do this in multiple places, such as:

```julia
run_git(cmd) = run(`git -c "user.name=Your Name" -c "user.email=Your Email" $(cmd)`)

git_clone(url::AbstractString, local_path::AbstractString) = run_git("clone $(url) $(local_path)")
```

something like this?

mattBrzezinski:
> > We'll need to pass e.g. `-c "user.name=Your Name" -c "user.email=Your email"` to every Git command. This allows us to provide the user name and user email to Git without any permanent side effects.
> 
> What do we think would be the best way to save "Your Name" and "Your email" to use it in multiple places? Would we make the user set a few ENVVARS with those values? Maybe have the user put `set_git_config(name, email)` and then save that somewhere to use later? Maybe raise an error if they are not set?
> 
> Potentially it might be nice to make some sort of command to call git so you don't need to do this in multiple places, such as:
> 
> ```julia
> run_git(cmd) = run(`git -c "user.name=Your Name" -c "user.email=Your Email" $(cmd)`)
> 
> git_clone(url::AbstractString, local_path::AbstractString) = run_git("clone $(url) $(local_path)")
> ```
> 
> something like this?

AFAIK this actually isn't needed for all git commands. In my test case it was needed for the `commit` command, but none of the others (clone, branch, etc).

DilumAluthge:
In theory we should hard-code the user and email as specific to CompatHelper. E.g. the user name is `CompatHelper` and the email is `compathelper-noreply@julialang.org` or whatever.

mattBrzezinski:
> In theory we should hard-code the user and email as specific to CompatHelper. E.g. the user name is `CompatHelper` and the email is `compathelper-noreply@julialang.org` or whatever.

I believe we can set these as the environment variables `GIT_COMMITTER_NAME` and `GIT_COMMITTER_EMAIL` in the repo settings and have them work? I do not have permissions to do this, but if you want to give it a shot I can update the code and we can test this out!

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

DilumAluthge:
Not sure I follow. What I'm saying is that whenever CompatHelper makes a PR to any repo, it should use that user name and email. So it's not a repo-specific setting.

So we could have global constants in the CompatHelper module:
```julia
const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
```

And then whenever the CompatHelper codebase needs, it can do e.g. `git -c "user.name=$(COMPATHELPER_GIT_COMMITTER_NAME)" -c "user.email=$(COMPATHELPER_GIT_COMMITTER_EMAIL)" commit ...`.

mattBrzezinski:
> Not sure I follow. What I'm saying is that whenever CompatHelper makes a PR to any repo, it should use that user name and email. So it's not a repo-specific setting.
> 
> So we could have global constants in the CompatHelper module:
> 
> ```julia
> const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
> const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
> ```
> 
> And then whenever the CompatHelper codebase needs, it can do e.g. `git -c "user.name=$(COMPATHELPER_GIT_COMMITTER_NAME)" -c "user.email=$(COMPATHELPER_GIT_COMMITTER_EMAIL)" commit ...`.

Ah right, sorry I was a bit confused here in what we were chatting about. This makes sense, I'll make the changes in the next commit.

mattBrzezinski:
> Not sure I follow. What I'm saying is that whenever CompatHelper makes a PR to any repo, it should use that user name and email. So it's not a repo-specific setting.
> 
> So we could have global constants in the CompatHelper module:
> 
> ```julia
> const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
> const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
> ```
> 
> And then whenever the CompatHelper codebase needs, it can do e.g. `git -c "user.name=$(COMPATHELPER_GIT_COMMITTER_NAME)" -c "user.email=$(COMPATHELPER_GIT_COMMITTER_EMAIL)" commit ...`.

Made these changes in the latest commit. Once I introduce a function for `git_commit()` I'll refactor a bit more as we go.

fchorney:
> Not sure I follow. What I'm saying is that whenever CompatHelper makes a PR to any repo, it should use that user name and email. So it's not a repo-specific setting.
> 
> So we could have global constants in the CompatHelper module:
> 
> ```julia
> const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
> const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
> ```
> 
> And then whenever the CompatHelper codebase needs, it can do e.g. `git -c "user.name=$(COMPATHELPER_GIT_COMMITTER_NAME)" -c "user.email=$(COMPATHELPER_GIT_COMMITTER_EMAIL)" commit ...`.

Would we also be able to override this with ENVVARS of some sort. If we want to use this in our Private GitLab repo eventually, I would like the commits/etc to be signed to a specific user in our organization (a bot, but a user that exists in our org none the less).

DilumAluthge:
> > Not sure I follow. What I'm saying is that whenever CompatHelper makes a PR to any repo, it should use that user name and email. So it's not a repo-specific setting.
> > So we could have global constants in the CompatHelper module:
> > ```julia
> > const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
> > const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And then whenever the CompatHelper codebase needs, it can do e.g. `git -c "user.name=$(COMPATHELPER_GIT_COMMITTER_NAME)" -c "user.email=$(COMPATHELPER_GIT_COMMITTER_EMAIL)" commit ...`.
> 
> Would we also be able to override this with ENVVARS of some sort. If we want to use this in our Private GitLab repo eventually, I would like the commits/etc to be signed to a specific user in our organization (a bot, but a user that exists in our org none the less).

Yeah that seems easy enough to me. We could have environment variables named `COMPATHELPER_GIT_COMMITTER_NAME` etc. If the environment variable exists (and has length >=1), we use that value. Else, we use the default value from the global constant.

fchorney (approved):

PR #335, opened 2021-06-07
[v3-dev] Some style tweaks: replace `using Foo` with `using Foo: Foo`, etc.

DilumAluthge:
I'm not a huge fan of `using Foo` in package code, because it brings all of `Foo`'s exports into scope, and it can be hard to tell (when reading the package code) where certain functions and types come from.

Instead, I prefer to use either `import Foo` or `using Foo: Foo`. The Blue Style guide discourages the former, and I want to stay compliant with Blue Style, so let us use the latter.


DilumAluthge:
bors try

DilumAluthge:
bors try-

DilumAluthge:
bors merge

PR #339, opened 2021-06-07
NewVersions utility functions

mattBrzezinski:
I was originally going to re-write the [new_versions.jl](https://github.com/JuliaRegistries/CompatHelper.jl/blob/0afb9df0164b9389ffbd444978b7e9c2e61fa5b3/src/new_versions.jl) file, but realized there are too many smaller functions and it would become an eye-sore to read through.

Instead, here's a smaller PR for some utility functions that will be used in a future PR.

### DIFF
--- a/src/utilities/new_versions.jl
+++ b/src/utilities/new_versions.jl
@@ -40,7 +40,7 @@ function subdir_string(subdir::AbstractString)
 end
 
 function pr_info(
-    version_verbatim::Nothing,
+    compat_entry_verbatim::Nothing,
     name::AbstractString,
     compat_entry_for_latest_version::AbstractString,
     compat_entry::AbstractString,
@@ -64,7 +64,7 @@ function pr_info(
 end
 
 function pr_info(
-    version_verbatim::AbstractString,
+    compat_entry_verbatim::AbstractString,
     name::AbstractString,
     compat_entry_for_latest_version::AbstractString,
     compat_entry::AbstractString,
@@ -78,7 +78,7 @@ function pr_info(
     """
 
     new_pr_body = m"""
-        This pull request changes the compat entry for the `$(name)` package from `$(version_verbatim)` to `$(compat_entry)` $(subdir_string). $(pr_body_keep_or_drop)
+        This pull request changes the compat entry for the `$(name)` package from `$(compat_entry_verbatim)` to `$(compat_entry)` $(subdir_string). $(pr_body_keep_or_drop)
         Note: I have not tested your package with this new compat entry.
         It is your responsibility to make sure that your package tests pass before you merge this pull request.
     """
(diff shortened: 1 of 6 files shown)


mattBrzezinski:
bors try

src/utilities/new_versions.jl
```diff
@@ -0,0 +1,122 @@
...
+    version_verbatim::Nothing,
+    name::AbstractString,
+    compat_entry_for_latest_version::AbstractString,
+    compat_entry::AbstractString,
+    subdir_string::AbstractString,
+    pr_body_keep_or_drop::AbstractString,
+    pr_title_parenthetical::AbstractString,
+)
+    new_pr_title = string(
+        "CompatHelper: add new compat entry for ",
```
fchorney, line 50: Dunno if we want to add another dep just for this, but https://github.com/invenia/MultilineStrings.jl might be nice to make these strings a little nicer.

test/utilities/new_versions.jl
```diff
@@ -0,0 +1,122 @@
...
+    for case in cases
+        bump_specifier, verbatim, expected = case
+        @test CompatHelper.skip_equality_specifiers(bump_specifier, verbatim) == expected
+    end
+
+    # If bump_compat_containing_equality_specifier is set to true, always return back false
+    for case in cases
+        bump_specifier, verbatim, expected = case
+        @test CompatHelper.skip_equality_specifiers(!bump_specifier, verbatim) == false
+    end
```
fchorney, line 110: could this comment and the test go in the other for loop? Seems fine either way but personally feel like this shouldn't be split into 2 loops

test/utilities/new_versions.jl
```diff
@@ -0,0 +1,122 @@
...
+
+            @test result == expected
+        end
+    end
+end
+
+@testset "compat_version_number -- $(vn)" for (vn, expected) in [
+    (VersionNumber("1.0.0"), "1"),
+    (VersionNumber("1.1.1"), "1"),
+    (VersionNumber("1.1.1"), "1"),
```
fchorney, line 70: this test is duplicated

src/utilities/new_versions.jl
DilumAluthge: Looks like MultilineStrings.jl has no dependencies, right?

I'd be fine adding it, since it's so lightweight. These multi-line strings really are annoying to deal with.

mattBrzezinski: My only issue w/ MLS is that VSCode doesn't understand these variables are being used inside of the macro :/

Screen Shot 2021-06-08 at 9 43 42 AM

DilumAluthge: I don't feel strongly either way.

mattBrzezinski:
bors try

fchorney (approved):
I notice the multiline strings, are slightly different than they used to be, but I don't have any strong opinions on them. @DilumAluthge any comments? 

LGTM otherwise

DilumAluthge:
`version_verbatim` should be `compat_entry_verbatim` instead?

DilumAluthge:
Yeah. Since that variable refers to `[compat]` entry strings, right?

Otherwise LGTM.

mattBrzezinski:
Bors is down?

DilumAluthge:
Yeah, looks like it.

PR #343, opened 2021-06-22
Add main function and tests

fchorney:
Adding the `main.jl` file, and tests for it.

### DIFF
--- a/src/utilities/new_versions.jl
+++ b/src/utilities/new_versions.jl
@@ -47,9 +47,10 @@ function pr_info(
     subdir_string::AbstractString,
     pr_body_keep_or_drop::AbstractString,
     pr_title_parenthetical::AbstractString,
+    pr_title_prefix::String,
 )
     new_pr_title = m"""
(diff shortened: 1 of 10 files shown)


src/utilities/new_versions.jl
```diff
@@ -162,7 +164,8 @@ end
 
 function continue_with_pr(dep::DepInfo, bump_compat_containing_equality_specifier::Bool)
     # Determine if we need to make a new PR
-    if !isnothing(dep.version_spec) && dep.latest_version in dep.version_spec
+    if (!isnothing(dep.version_spec) && !isnothing(dep.latest_version)) &&
+       dep.latest_version in dep.version_spec
```
fchorney, line 168: looks like this change is required because `nothing in VersionSpec` doesn't have a method for it.

fchorney:
bors try

mattBrzezinski (approved):
LGTM! After this I suppose we test it, then add in the GitLab configurability stuff.

mattBrzezinski:
bors try

PR #349, opened 2021-07-08
Add Full GitLab Support

fchorney:
Adding in GitLab support as I test it on our system.

### DIFF
--- a/src/main.jl
+++ b/src/main.jl
@@ -19,19 +19,7 @@ function main(
     unsub_from_prs=false,
     cc_user=false,
 )
-    if ci_cfg isa GitHubActions
-        token = GitHub.Token(github_token())
-        api = GitHub.GitHubAPI(; token=token, url=hostname_for_api)
-        repo, _ = @mock GitForge.get_repo(api, github_repository())
-    elseif ci_cfg isa GitLabCI
-        token = GitLab.PersonalAccessToken(gitlab_token())
-        api = GitLab.GitLabAPI(; token=token, url=hostname_for_api)
-        repo, _ = @mock GitForge.get_repo(api, gitlab_repository())
-    else
-        err = "Unknown CI Config"
-        @error(err)
-        throw(ErrorException(err))
-    end
+    api, repo = get_api_and_repo(ci_cfg, hostname_for_api)
 
     for subdir in subdirs
         deps = @mock get_project_deps(
--- a/src/main.jl
+++ b/src/main.jl
@@ -16,6 +16,7 @@ function main(
     bump_compat_containing_equality_specifier=true,
     pr_title_prefix::String="",
     include_jll::Bool=false,
+    unsub_from_prs=false,
 )
     if ci_cfg isa GitHubActions
         token = GitHub.Token(github_token())
@@ -50,6 +51,7 @@ function main(
                 env=env,
                 bump_compat_containing_equality_specifier=bump_compat_containing_equality_specifier,
                 pr_title_prefix=pr_title_prefix,
+                unsub_from_prs=unsub_from_prs,
             )
         end
     end
--- a/src/main.jl
+++ b/src/main.jl
@@ -17,6 +17,7 @@ function main(
     pr_title_prefix::String="",
     include_jll::Bool=false,
     unsub_from_prs=false,
+    cc_user=false,
 )
     if ci_cfg isa GitHubActions
         token = GitHub.Token(github_token())
@@ -52,6 +53,7 @@ function main(
                 bump_compat_containing_equality_specifier=bump_compat_containing_equality_specifier,
                 pr_title_prefix=pr_title_prefix,
                 unsub_from_prs=unsub_from_prs,
+                cc_user=cc_user,
             )
         end
     end
--- a/src/main.jl
+++ b/src/main.jl
@@ -21,6 +21,10 @@ function main(
         token = GitHub.Token(github_token())
         api = GitHub.GitHubAPI(; token=token, url=hostname_for_api)
         repo, _ = @mock GitForge.get_repo(api, github_repository())
+    elseif ci_cfg isa GitLabCI
+        token = GitLab.PersonalAccessToken(gitlab_token())
+        api = GitLab.GitLabAPI(; token=token, url=hostname_for_api)
+        repo, _ = @mock GitForge.get_repo(api, gitlab_repository())
     else
         err = "Unknown CI Config"
         @error(err)
--- a/src/utilities/ci.jl
+++ b/src/utilities/ci.jl
@@ -3,30 +3,36 @@ abstract type CIService end
 struct GitHubActions <: CIService
     username::String
     email::String
+
+    function GitHubActions(
+        username="github-actions[bot]",
+        email="41898282+github-actions[bot]@users.noreply.github.com",
+    )
+        return new(username, email)
+    end
 end
 
 struct GitLabCI <: CIService
     username::String
     email::String
-end
 
-function GitHubActions()
-    return GitHubActions(
-        "github-actions[bot]", "41898282+github-actions[bot]@users.noreply.github.com"
+    function GitLabCI(
+        username="gitlab-ci[bot]",
+        email="gitlab-ci[bot]@gitlab.com",
     )
-end
-
-function GitLabCI()
-    return GitLabCI("gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com")
+        return new(username, email)
+    end
 end
 
 api_hostname(::GitHubActions) = "https://api.github.com"
 api_hostname(::GitLabCI) = "https://gitlab.com/api/v4"
+
 clone_hostname(::GitHubActions) = "github.com"
 clone_hostname(::GitLabCI) = "gitlab.com"
 
 ci_repository(::GitHubActions, env::AbstractDict=ENV) = env["GITHUB_REPOSITORY"]
 ci_repository(::GitLabCI, env::AbstractDict=ENV) = env["CI_PROJECT_PATH"]
+
 ci_token(::GitHubActions, env::AbstractDict=ENV) = env["GITHUB_TOKEN"]
 ci_token(::GitLabCI, env::AbstractDict=ENV) = env["GITLAB_TOKEN"]
 
--- a/src/utilities/ci.jl
+++ b/src/utilities/ci.jl
@@ -30,7 +30,6 @@ struct GitLabCI <: CIService
 end
 
 function GitLabCI()
-    # TODO: Not sure what to actually put here
     return GitLabCI("gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com")
 end
 
--- a/src/utilities/ci.jl
+++ b/src/utilities/ci.jl
@@ -14,8 +14,15 @@ end
 api_hostname(::GitHubActions) = "https://api.github.com"
 clone_hostname(::GitHubActions) = "github.com"
 
-github_repository(env::AbstractDict=ENV) = env["GITHUB_REPOSITORY"]
-github_token(env::AbstractDict=ENV) = env["GITHUB_TOKEN"]
+ci_repository(::GitHubActions, env::AbstractDict=ENV) = env["GITHUB_REPOSITORY"]
+ci_token(::GitHubActions, env::AbstractDict=ENV) = env["GITHUB_TOKEN"]
+
+function get_api_and_repo(ci::GitHubActions, hostname_for_api::AbstractString)
+    token = GitHub.Token(ci_token(ci))
+    api = GitHub.GitHubAPI(; token=token, url=hostname_for_api)
+    repo, _ = @mock GitForge.get_repo(api, ci_repository(ci))
+    return api, repo
+end
 
 struct GitLabCI <: CIService
     username::String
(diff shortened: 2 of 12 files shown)


fchorney:
bors try

fchorney:
bors try

fchorney:
bors try

src/main.jl
```diff
@@ -18,11 +18,17 @@ function main(
...
 )
     if ci_cfg isa GitHubActions
         token = GitHub.Token(github_token())
         api = GitHub.GitHubAPI(; token=token, url=hostname_for_api)
         repo, _ = @mock GitForge.get_repo(api, github_repository())
+    elseif ci_cfg isa GitLabCI
+        token = GitLab.PersonalAccessToken(gitlab_token())
+        api = GitLab.GitLabAPI(; token=token, url=hostname_for_api)
+        repo, _ = @mock GitForge.get_repo(api, gitlab_repository())
     else
```
oxinabox, line 32: Maybe pull these 3 branchs out into little methods that return the `api` and the `repo`?

src/utilities/ci.jl
```diff
@@ -16,9 +17,29 @@ clone_hostname(::GitHubActions) = "github.com"
...
+    return GitLabCI(
+        "gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com"
+    )
+end
+
+api_hostname(::GitLabCI) = "https://gitlab.com/api/v4"
+clone_hostname(::GitLabCI) = "gitlab.com"
+
+gitlab_repository(env::AbstractDict=ENV) = env["CI_PROJECT_PATH"]
+gitlab_token(env::AbstractDict=ENV) = env["GITLAB_TOKEN"]
```
oxinabox, line 36:
```suggestion
repository(::GitLabCI, env::AbstractDict=ENV) = env["CI_PROJECT_PATH"]
token(:GitLabCI, env::AbstractDict=ENV) = env["GITLAB_TOKEN"]
```
and similar changed to the `GitHubCI` bit?

src/utilities/ci.jl
```diff
@@ -16,9 +17,29 @@ clone_hostname(::GitHubActions) = "github.com"
...
+    username::String
+    email::String
+end
+
+function GitLabCI()
+    # TODO: Not sure what to actually put here
+    return GitLabCI(
+        "gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com"
+    )
+end
```
oxinabox, line 30:
```suggestion
# TODO: Not sure what to actually put here
GitLabCI() = GitLabCI("gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com")
```

src/utilities/new_versions.jl
```diff
@@ -110,15 +114,17 @@ function create_new_pull_request(
     title::AbstractString,
     body::AbstractString,
 )
+    repo_owner = String(rsplit(repo.path_with_namespace, "/"; limit=2)[1])
```
oxinabox, line 117: Should this be a helper that takes ` repo` ?

src/utilities/new_versions.jl
```diff
@@ -166,24 +172,30 @@ function continue_with_pr(dep::DepInfo, bump_compat_containing_equality_specifie
     # Determine if we need to make a new PR
     if (!isnothing(dep.version_spec) && !isnothing(dep.latest_version)) &&
        dep.latest_version in dep.version_spec
+        # TODO: This seems too noisy, is there a better fix?
```
oxinabox, line 175: Using LoggingExtras to filter out messages you don't want at the end user level.
Maybe add a `group` arg here so you can filter them more easily

src/utilities/new_versions.jl
```diff
@@ -280,24 +295,69 @@ function make_pr_for_new_version(
...
+                cc_user && cc_trigger_user(forge, repo, new_pr; env=env)
+                unsub_from_prs && unsub_from_pr(forge, new_pr)
             end
         end
     end
 
     return nothing
 end
 
+function cc_trigger_user(
```
oxinabox, line 321: probably clearer to say 
```suggestion
function cc_mention_user(
```
?
I think both github and gitlab use this term?

src/utilities/new_versions.jl
```diff
@@ -280,24 +295,69 @@ function make_pr_for_new_version(
...
+    return @mock GitForge.create_pull_request_comment(
+        api,
+        repo.id,
+        pr.iid;
+        body=body,
+    )
+end
+
+function unsub_from_pr(api::GitHub.GitHubAPI, pr::GitHub.PullRequest)
+    # Not implemented for GitHub
```
oxinabox, line 353: Probably print a warning?
Or even let GitForge throw an error?

src/utilities/new_versions.jl
```diff
@@ -280,24 +295,69 @@ function make_pr_for_new_version(
...
+end
+
+function cc_trigger_user(
+    api::GitLab.GitLabAPI, repo::GitLab.Project, pr::GitLab.MergeRequest;
+    env=ENV,
+)
+    username = env["GITLAB_USER_LOGIN"]
+    body = "cc @$username"
+
+    return @mock GitForge.create_pull_request_comment(
```
oxinabox, line 344: why are these different?
Point of GitForge is that they should have common API

src/utilities/new_versions.jl
```diff
@@ -280,24 +295,69 @@ function make_pr_for_new_version(
...
+        pr.id;
+        body=body,
+    )
+end
+
+function cc_trigger_user(
+    api::GitLab.GitLabAPI, repo::GitLab.Project, pr::GitLab.MergeRequest;
+    env=ENV,
+)
+    username = env["GITLAB_USER_LOGIN"]
```
oxinabox, line 341: I wonder if we should add a bunch of env-var helpers to GitForge?

test/runtests.jl
```diff
@@ -28,9 +29,9 @@ Aqua.test_all(CompatHelper; ambiguities=false)
     @test occursin(Regex("\\sversion = \"$(major_version)\"\n"), workflow_filecontents)
     @test length(findall(r"version[\s]*?=", workflow_filecontents)) == 1
 end
-
 include("patches.jl")
 
+cwd = pwd()
```
oxinabox, line 34: This isn't used
```suggestion
```

src/utilities/ci.jl
fchorney: ah yeah that makes sense, with the other change to that if block above, I think this would be nicer.

src/utilities/new_versions.jl
fchorney: Unfortunately gitforge isn't perfect in this regard. Some API calls for gitlab specifically need an owner and repo as strings, while in gitlab require the project id. This causes some situations where you need to separate how github and gitlab do calls. This also appears in the fact that the naming of classes/variables/kwargs is different for the "same" command on gitlab/hub

test/runtests.jl
fchorney: it's used at the top of the `tests/utilities/new_versions.jl` file to avoid an error where we are cd'd into a tempdir that gets deleted. There is possibly a better solution, but would require a rework of many tests.

src/utilities/new_versions.jl
oxinabox: We should fix GitForge to take a higher level object as a argument, that contains as needed both owner/repo, and project id.

test/runtests.jl
oxinabox:
```suggestion
const cwd = pwd()
```

fchorney: the const is fine, it needs to grab the pwd before the tests run.

src/utilities/new_versions.jl
fchorney: This seems very out of scope for right now. I think it would be best to make an issue, but this seems like it would be a big change to overhaul gitforge to properly be more generic

fchorney: that being said, I think it would be nice to continue to show love to gitforge and contribute to make it better

src/utilities/new_versions.jl
fchorney: Possibly? I'm on the fence about factoring out a single line that happens only once, as I don't think it's going to be used anywhere else.

mattBrzezinski: I think if we're doing this once it's fine. If we end up needing similar code multiple times we should then turn it into a helper function.

src/utilities/new_versions.jl
fchorney: hmm interesting, I'll look into this thanks!

test/main.jl
```diff
@@ -26,4 +26,28 @@
             end
         end
     end
+
+    @testset "successful run GitLab" begin
+        mktempdir() do tmpdir
+            cd(tmpdir)
```
mattBrzezinski, line 32: Should probably just match what I did here,

https://github.com/JuliaRegistries/CompatHelper.jl/pull/350/files#diff-21e3ffd1ceccd2a142495277ae65b3e5bd5569c7ef00b4573757281e90bb2a11L10

fchorney: I think overall there is a bit of an issue with the tests `cd`ing into tmp dirs and then just staying there after the tempdirs disappear. I think the tests need a bit of an overhaul to clean that up a bit.

mattBrzezinski: It's more-so that we don't `cd(original_dir)` here that was causing me issues with integration testing afterwards.

I did clean up *some* of these in my PR.

src/utilities/new_versions.jl
fchorney: I think this is a good idea, and would warrant an issue on GitForge. Also a little bit out of scope imo, but would be nice to set up pre-defined CI ENVVARs and an easy way to grab them without knowing the actual var name, like `e = ForgeCIEnv(::GitLabAPI)` `username(api, e)`, something like that in any case

fchorney: Currently GitForge doesn't necessarily concern itself with CI's, but it probably should

src/utilities/new_versions.jl
fchorney: I think I'll open an issue for this, and just let the output happen in this PR for now.

fchorney:
bors try

src/utilities/ci.jl
```diff
@@ -13,12 +14,49 @@ end
...
+    token = GitHub.Token(ci_token(ci))
+    api = GitHub.GitHubAPI(; token=token, url=hostname_for_api)
+    repo, _ = @mock GitForge.get_repo(api, ci_repository(ci))
+    return api, repo
+end
+
+struct GitLabCI <: CIService
+    username::String
+    email::String
+end
```
mattBrzezinski, line 30: Maybe move this to the top of the file

src/utilities/ci.jl
```diff
@@ -13,12 +14,49 @@ end
...
+
+function GitLabCI()
+    return GitLabCI("gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com")
+end
+
+api_hostname(::GitLabCI) = "https://gitlab.com/api/v4"
+clone_hostname(::GitLabCI) = "gitlab.com"
```
mattBrzezinski, line 37: Place these closer to the `::GitHubActions` ones maybe?

src/utilities/ci.jl
```diff
@@ -13,12 +14,49 @@ end
...
+    token = GitLab.PersonalAccessToken(ci_token(ci))
+    api = GitLab.GitLabAPI(; token=token, url=hostname_for_api)
+    repo, _ = @mock GitForge.get_repo(api, ci_repository(ci))
+    return api, repo
+end
+
+function get_api_and_repo(ci::Any, hostname_for_api::AbstractString)
+    err = "Unknown CI Config: $(typeof(ci))"
+    @error(err)
+    throw(ErrorException(err))
```
mattBrzezinski, line 52: I don't understand what we're doing here?

src/utilities/new_versions.jl
```diff
@@ -219,7 +227,8 @@ function make_pr_for_new_version(
     brand_new_compat = new_compat_entry(
         entry_type, dep.version_verbatim, compat_entry_for_latest_version
     )
-    new_pr_title, new_pr_body = pr_info(
+    new_pr_title,
+    new_pr_body = pr_info(
```
mattBrzezinski, line 231: Keep on same line?

src/utilities/new_versions.jl
```diff
@@ -280,24 +289,68 @@ function make_pr_for_new_version(
...
+        body=body,
+    )
+end
+
+function unsub_from_pr(api::GitHub.GitHubAPI, pr::GitHub.PullRequest)
+    return GitForge.unsubscribe_from_pull_request(api, pr.repo.id, pr.id)
+end
+
+function unsub_from_pr(api::GitLab.GitLabAPI, pr::GitLab.MergeRequest)
+    return @mock GitForge.unsubscribe_from_pull_request(api, pr.project_id, pr.iid)
```
mattBrzezinski, line 351: Just confirming `pr.iid` is a thing and not a typo of `pr.id`

test/main.jl
```diff
@@ -4,30 +4,55 @@
...
+                        CompatHelper.main()
+                    end
+                end
+            end
+        end
+    end
+
+    @testset "successful run GitLab" begin
+        mktempdir() do tmpdir
+            cd(tmpdir) do
```
mattBrzezinski, line 39: Personally, I think the suggestion below is a *bit* nicer. It doesn't indent as much and lowers the chance of over running line length.

```suggestion
        tmpdir = mktempdir()
        cd(tmpdir) do
```

test/utilities/new_versions.jl
```diff
@@ -289,66 +351,66 @@ end
...
+                    git_push_patch,
+                    gh_pr_patch,
+                    make_clone_https_patch(tmpdir),
+                    make_clone_ssh_patch(tmpdir),
+                    decode_pkey_patch,
+                ]
+
+                apply(patches) do
+                    # HTTPS
+
```
mattBrzezinski, line 378:
```suggestion
```

mattBrzezinski (approved):

src/utilities/new_versions.jl
fchorney: yup, iid is internal id

src/utilities/new_versions.jl
fchorney: I'd like to, but the formatter keeps moving it :/

src/utilities/ci.jl
fchorney: Printing an error to the log and then throwing an exception. This will only get called if we run `get_api_and_repo` with a `CIService` that doesn't properly implement this function.

src/utilities/new_versions.jl
fchorney: oh hmm nevermind, the formatter says it should be on the same line now

fchorney:
bors try

src/utilities/ci.jl
```diff
@@ -4,21 +4,59 @@ struct GitHubActions <: CIService
     username::String
     email::String
 end
+
+struct GitLabCI <: CIService
+    username::String
+    email::String
+end
```
mattBrzezinski, line 11: I can't comment around this for the `GitHubActions` :/ 
Note: This doesn't allow you to do `GitLabCI("username", "email")` though.

Not sure if that's functionality we want to keep or not /shrug

```suggestion
struct GitLabCI <: CIService
    username::String
    email::String
    
    function GitLabCI()
        return new("gitlab-ci[bot]", "gitlab-ci[bot]@gitlab.com")
    end
end
```

test/main.jl
mattBrzezinski: Yeah not a big deal, I don't think when it's destroyed matters much. We can leave as is and if it gets too lengthy (side-ways) we can refactor.

fchorney:
bors try

PR #355, opened 2021-07-20
Refactored API Hostname and Clone Hostname into CIService struct

mattBrzezinski:


mattBrzezinski:
bors try

fchorney (approved):
afaict this is good 2 go

PR #356, opened 2021-07-20
Minor cleanup, whitespace, moving functions around,etc

mattBrzezinski:
Based off of `MB/ci_service_refactor`, point to `v3-dev` later.


src/utilities/git.jl
```diff
@@ -1,6 +1,9 @@
 const COMPATHELPER_GIT_COMMITTER_NAME = "CompatHelper Julia"
 const COMPATHELPER_GIT_COMMITTER_EMAIL = "compathelper_noreply@julialang.org"
 
+git_add() = run(`git add -A .`)
```
fchorney, line 4: ah yes good catch. I had add doing a bunch of other stuff so I put it in a function and didn't notice it was small enough to put into a one liner once I removed the other stuff

fchorney (approved):

PR #358, opened 2021-07-21
Bump Version of Package

### LINKED ISSUE #132: Opt-in feature: bump patch component of version number

fchorney:
It seems typical to bump the package version when adding dependency compat values. I have set this to false by default but if we want it true by default I can change it.

If the package is >= 1.0 then the minor value will be bumped and the patch will go to 0. If the package is < 1.0 then only the patch value will be bumped.

I had a request to ignore the bumping of version values if a pre-release strings exists in the version number such as `-DEV`. Let me know if this needs to be relaxed or if we want a different rule for it.

Note: This was made against the v3-dev branch. 

closes #132

### DIFF
--- a/src/utilities/new_versions.jl
+++ b/src/utilities/new_versions.jl
@@ -275,12 +275,13 @@ function make_pr_for_new_version(
             new_branch_name = "compathelper/new_version/$(get_random_string())"
             git_branch(new_branch_name; checkout=true)
 
-            # Add new compat entry to project.toml and write it out
-            add_compat_entry(
+            # Add new compat entry to project.toml, bump the version if needed,
+            # and write it out
+            modify_project_toml(
                 dep.package.name,
                 joinpath(tmpdir, LOCAL_REPO_NAME, subdir),
-                brand_new_compat;
-                bump_version=bump_version,
+                brand_new_compat,
+                bump_version,
             )
             git_add()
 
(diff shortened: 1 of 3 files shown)


src/utilities/new_versions.jl
```diff
@@ -340,21 +342,51 @@ function unsub_from_pr(api::GitLab.GitLabAPI, pr::GitLab.MergeRequest)
...
+function bump_package_version!(project::Dict)
+    version = VersionNumber(project["version"])
+
+    # Only bump the version if prerelease is empty
+    !isempty(version.prerelease) && return nothing
+
+    # Bump minor if version > 1.0, else bump patch
+    if version.major >= 1
+        version = VersionNumber(
+            version.major, version.minor + 1, 0, version.prerelease, version.build
```
mattBrzezinski, line 374: I don't think prerelease or build are used in compat entries, so they're not needed?
I feel like keeping the same build number especially is wrong, as that would not really exist?

fchorney: This updates the package version, not the compat version. That being said, maybe this shouldn't be in the same function that updates the compat version

fchorney:
bors try

DilumAluthge:
> I have set this to false by default but if we want it true

Yeah I think `false` by default is good.

Would it make sense to just always bump the patch version? I don't think a compat bump needs a minor version bump in a post 1.0 packages, does it?

fchorney:
> Would it make sense to just always bump the patch version? I don't think a compat bump needs a minor version bump in a post 1.0 packages, does it?

According to [ColPrac ](https://github.com/SciML/ColPrac) it seems that it should. It has this section about changing compat entries:

### Changing dependency compatibility

*   Generally changing dependency compatibility should be a non-breaking feature.
    *   i.e. pre-1.0 change patch version number, post-1.0 change the minor version number.
    *   For instance, adding or removing compatibility with a particular __version__ of a current dependency, which may or may not require internal code changes.
    *   This also applies when adding or removing packages as dependencies.
    *   The new feature in question is the ability to use with a different set of packages.
*   Changing a dependency to resolve a bug is a bug-fix.
    *   i.e. pre/post-1.0 change patch version number.
    *   For instance, if a bug in a downstream dependency is causing a problem in your package restricting compat to not allow that version would be a bug-fix.
*   Changing compatibility with dependencies **may** be a breaking release, if it breaks the user-facing interface.
    That is to say if the dependency’s API leaks into your API.
    There are three ways that this can happen:
    *   Reexporting a function that has changed.
    *   Returning an object of a type that’s behaviour has changed.
    *   Subtyping an object that has changed.

fchorney:
bors try

fchorney:
bors try

DilumAluthge:
> I had a request to ignore the bumping of version values if a pre-release strings exists in the version number such as `-DEV`.

I think this rule sounds good.

fchorney:
Any other comments or does this seem good to go?

mattBrzezinski (approved):

PR #359, opened 2021-07-22
Added documentation

mattBrzezinski:
@fchorney if you don't mind adding in documentation for `GitHub/GitLab` setup instructions?


README.md
```diff
@@ -19,6 +21,21 @@
...
+| Name | Description |
+| ---- | ----------- |
+| CI_PROJECT_PATH | Name on GitLab of the `organization/repo`, provided by default on GitLab |
+| GIT_COMMITTER_NAME | Name to associate commits with |
+| GIT_COMMITTER_EMAIL | Email to associate commits with |
+| GITHUB_ACTOR | Users on GitHub to CC on pull requests |
+| GITHUB_REPOSITORY | Name on GitHub of the `organization/repo`, provided by default on GitHub |
+| GITHUB_TOKEN | GitHub Access Token, provided by default by GitHub Actions |
+| GITLAB_TOKEN | GitLab Access Token, provided by default by GitLab CI |
+| GITLAB_USER_LOGIN | Users on GitLab to CC on pull requests |
```
fchorney, line 41: This is provided by default by GitLab CI

README.md
```diff
@@ -19,6 +21,21 @@
...
-Please see the [documentation](https://JuliaRegistries.github.io/CompatHelper.jl/dev/).
+| Name | Description |
+| ---- | ----------- |
+| CI_PROJECT_PATH | Name on GitLab of the `organization/repo`, provided by default on GitLab |
+| GIT_COMMITTER_NAME | Name to associate commits with |
+| GIT_COMMITTER_EMAIL | Email to associate commits with |
+| GITHUB_ACTOR | Users on GitHub to CC on pull requests |
+| GITHUB_REPOSITORY | Name on GitHub of the `organization/repo`, provided by default on GitHub |
+| GITHUB_TOKEN | GitHub Access Token, provided by default by GitHub Actions |
+| GITLAB_TOKEN | GitLab Access Token, provided by default by GitLab CI |
```
fchorney, line 40: I don't think either of these are provided by default? I thought these were the ones we need to make secrets for.

mattBrzezinski: https://cli.github.com/manual/gh_help_environment

I am unsure of GL here. We needed to definitely provide that since we're launching off of GHA.

fchorney: hmm right. I guess it kind of depends on the permissions you have set for that user. 

GitLab has similar tokens such as `CI_JOB_TOKEN` (https://docs.gitlab.com/ee/ci/variables/predefined_variables.html) which creates a token for that particular job, but I don't think you can access the `master` Personal Access Key for the user running the pipeline/job.

Looks like this is what the job token gets you (https://docs.gitlab.com/ee/api/index.html#gitlab-cicd-job-token) which I don't think would be sufficient for making MR's to the project. I think for GL you might have to specify the access token in the secrets rather than relying on this. 

I might try some tests using the CI_JOB_TOKEN to see if that would work if you ran CH on GitLab in a similar way that we use it on GitHub.

fchorney: Also looking at the page you linked, I don't think those are provided by GitHub Actions, as at the top it says `Environment variables that can be used with gh`

fchorney: Here is the list of GitHub Pre-Defined environment variables https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

docs/src/index.md
```diff
@@ -14,70 +14,46 @@ Whenever one of your package's dependencies releases a new breaking version, Com
...
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
+        run: julia -e 'using CompatHelper; CompatHelper.main()'
 ```
+### Creating SSH Key
+If you use GitHub Actions to either test your packge using continuous integration, or build and deploy documentation you will need to create an SSH deploy key. If you wish to reuse an existing SSH key simplify modify the workflow above environment variable to use `COMPATHELPER_PRIV: ${{ secrets.DOCUMENTER_KEY }}`. Otherwise follow the below instructions to generate a new key:
+
+1. Generate a new SSH key
```
fchorney, line 43: might be good to explain/give and example of making a base-64 encoded key. I know on GitLab at least, it makes more sense to do that, not sure it really matters with github

docs/src/index.md
```diff
@@ -14,70 +14,46 @@ Whenever one of your package's dependencies releases a new breaking version, Com
...
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
+        run: julia -e 'using CompatHelper; CompatHelper.main()'
```
fchorney, line 38: I'm not sure if it makes more sense to use this PkgTemplates version or to use the version that exists in this repo. The one in this repo is possibly better because it pins it to the major version, but on the other hand people probably don't want to have to update all their packages whenever a major change comes out. @DilumAluthge any comments on this?

mattBrzezinski: I had taken this from what already existed, but can def change it up!

README.md
mattBrzezinski: Hmmm I can't seem to find it anymore, maybe I'm just blind. There's a default ENV VAR which provides the `GITHUB_TOKEN`, or something like that for making authenticated requests to the GH API so you don't get rate limited.

docs/src/index.md
mattBrzezinski: This is taken from the existing documentation. Is there really a point to storing as base64?

fchorney: at least in GitLab, you can't hide the ssh key from being outputted in it's entirety (if it is output somehow), if you store it in base 64 then it can be masked during logging and just show up as `***`. 

The existing documentation I believe also mentions base-64 encoded ssh keys, but doesn't really explain it

README.md
fchorney: Even in the integration tests we need to make `GITHUB_TOKEN` specifically from the secret https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/test/integration-tests.jl#L38, and in the unit tests we just mock all of the actual GitForge API calls, so not too sure which token or env var you might be thinking of.

docs/src/index.md
fchorney: yeah the one you have there is the one that exists in PkgTemplates, but the one that exists for this repo is here https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml

DilumAluthge:
Do we need to export `main`? Could we not export it, and just make it clear in the docstring that `CompatHelper.main` is part of the public API of the package?

docs/src/index.md
DilumAluthge: So, a couple of notes:

1. Regarding PkgTemplates, we (the CompatHelper.jl repo) are the upstream here, and thus we are the authoritative source. So the workflow file that exists in the CompatHelper.jl repo at `.github/workflows/CompatHelper.yml` is the authoritative workflow file. If the PkgTemplates version differs from our version, we should probably make a PR to edit the PkgTemplate version so that it exactly matches ours.
2. Yeah, it is intentional that our workflow file (`.github/workflows/CompatHelper.yml`) pins the major version. This means that, in the future, we (CompatHelper maintainers) can make breaking changes if we really want, and it won't break the whole ecosystem. It's a little inconvenient that people will have to change e.g. "2" to "3", but the alternative is that if we make a breaking change, hundreds of people start getting emails that their workflows are broken, which leads to a lot of irate users. In fact I did this once, and people were not happy 😂 .
3. Regarding the documentation, I'd like us to avoid repeating ourselves. That is, I don't want to have the same content in multiple places, because that means we have to update multiple locations. So the `.github/workflows/CompatHelper.yml` file should be the only place that content is defined. If we want to use that content elsewhere (e.g. in the docs), we should just include it from the `.github/workflows/CompatHelper.yml`. If you look how we do it in the docs on master:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/0afb9df0164b9389ffbd444978b7e9c2e61fa5b3/docs/src/index.md#L15-L28

fchorney: That makes sense to me. If it exists in the project we should just link it. I agree on using the version of CompatHelper.yml that we have in this project with the version specifically named. We should make an issue on PkgTemplates to get that changed.

README.md
```diff
@@ -19,6 +21,58 @@
...
+| Name | Description |
+| ---- | ----------- |
+| GIT_COMMITTER_NAME | Name to associate commits with [Default: CompatHelper Julia] |
+| GIT_COMMITTER_EMAIL | Email to associate commits with [Default: compathelper_noreply@julialang.org] |
+| COMPATHELPER_PRIV | Plain Text or Base64 Encoded SSH Public Key for git pushes and API calls |
+
+#### GitHub
+| Name | Description |
+| ---- | ----------- |
+| GITHUB_TOKEN | GitHub Access Token, user for GitHub API Requests |
```
fchorney, line 69:
```suggestion
| GITHUB_TOKEN | GitHub Access Token, user for GitHub API Requests, [provided by default on GitHub Actions](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) |
```

fchorney (approved):

PR #360, opened 2021-07-28
Main Returns the PRs that it creates

fchorney:
In some situations it would seem useful to do post processing after CompatHelper runs based on the PRs that were created during the run. I propose we just return a vector of PRs

### DIFF
--- a/src/main.jl
+++ b/src/main.jl
@@ -50,10 +50,12 @@ function main(
     cc_user=false,
     bump_version=false,
 )
+    generated_prs = Vector{Union{GitHub.PullRequest,GitLab.MergeRequest}}()
+
     api, repo = get_api_and_repo(ci_cfg)
 
     for subdir in subdirs
-        deps = @mock get_project_deps(
+        deps = get_project_deps(
             api,
             ci_cfg,
             repo;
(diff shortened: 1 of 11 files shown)


fchorney:
bors try

src/main.jl
```diff
@@ -79,8 +81,12 @@ function main(
                 cc_user=cc_user,
                 bump_version=bump_version,
             )
+
+            if !isnothing(pr)
```
mattBrzezinski, line 85: Do we have a standard for using `isnothing()` vs `=== nothing`?

src/utilities/new_versions.jl
```diff
@@ -247,6 +243,7 @@ function make_pr_for_new_version(
     end
 
     # Make a dir for our SSH PrivateKey which we will use only if it has been enabled
+    created_pr = nothing
```
mattBrzezinski, line 246: I've never really been a fan of this, but I feel like sometimes you need to?
I guess the alternative is to keep the `return nothing` at the end of the function, and a pre-mature `return new_pr`?

Doesn't really matter, idk if you have thoughts on it?

mattBrzezinski (approved):
LGTM!

src/utilities/new_versions.jl
fchorney: I think because of all the do blocks, you'd need to chain the returns up through every do block which would be more messy

src/main.jl
fchorney: I thought the current standard was to use `isnothing`

DilumAluthge: Someone told me than in 1.7, `isnothing` is defined in terms of `===`, so either should be fine.

Just don't use `==`, I think.

fchorney:
bors try

PR #363, opened 2021-08-10
create_pull_request should use id for GitLab

fchorney:
Looks like using owner/repo doesn't work well for projects in subgroups such as "group/subgroup/repo.jl", using just the numerical id instead is a better solution anyway.


mattBrzezinski (approved):

fchorney:
bors try

DilumAluthge:
bors merge

DilumAluthge:
bors merge

PR #364, opened 2021-08-10
Add flag for a short branch name

fchorney:
Hitting an issue where the branch names are too big for certain tests we are doing with AWS. This adds in a flag to use a shorter branch name.


fchorney:
This may not be needed, as we are looking into other solutions.

fchorney:
Nevermind, we found a different solution. Unless anyone else needs this I am going to close it.

DilumAluthge:
> Nevermind, we found a different solution. Unless anyone else needs this I am going to close it.

SGTM.

Just out of curiosity, what was your solution?

fchorney:
The issue was with role names, they must be 64 chars or smaller. We just opted to truncate them as that should be fine for our purposes

PR #365, opened 2021-08-11
Bug-fix for bump_package_version! when Project version is undefined

st--:
CompatHelper is in use for e.g. doc and test environments (additional dependencies) that do not define a version. This change ensures that CompatHelper won't error in that case.


fchorney:
oh good point! Thanks for catching this :)

mattBrzezinski (approved):

PR #374, opened 2021-08-12
Compare GH repos on fields rather than ==

mattBrzezinski:


DilumAluthge:
Should we also do the same thing for GitLab?

fchorney (approved):

PR #379, opened 2021-08-15
Recommend that most users use the new `compathelper-action` composite action

DilumAluthge:
This PR changes the officially recommended workflow file to use the new `compathelper-action` composite action: https://github.com/JuliaRegistries/compathelper-action

This PR should not be merged until we first tag a release of the `compathelper-action` action.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -7,26 +7,6 @@ jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - name: "Add the General registry via Git"
-        run: |
-          import Pkg
-          ENV["JULIA_PKG_SERVER"] = ""
-          Pkg.Registry.add("General")
-        shell: julia --color=yes {0}
-      - name: "Install CompatHelper"
-        run: |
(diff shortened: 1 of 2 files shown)


DilumAluthge:
bors try

SaschaMann:
Perhaps also add a GHA-warning to CompatHelper.jl so that people notice that they could/should switch more easily?

DilumAluthge:
> Perhaps also add a GHA-warning to CompatHelper.jl so that people notice that they could/should switch more easily?

We could, but:
1. How can we tell if someone's using the new composite action versus the old workflow file? Presumably there's no point in printing a warning if they've already switched to the composite action?
2. Do people actually look at the CompatHelper logs? My impression is that most people don't look at the logs, especially if the CompatHelper job isn't failing.

SaschaMann:
> Do people actually look at the CompatHelper logs? My impression is that most people don't look at the logs, especially if the CompatHelper job isn't failing.

Right, that's probably true. 1) could be done by searching the workflow dir but I didn't consider that people don't really look at these.

DilumAluthge:
bors try

PR #407, opened 2022-03-09
Make GHA workflow reusable

SaschaMann:
This allows people to use the workflow without copying its entirety, similarly to calling an action.

More info: https://docs.github.com/en/actions/using-workflows/reusing-workflows

Test run: https://github.com/julia-actions/Example.jl/runs/5474751558?check_suite_focus=true (for a scheduled test run, you might have to check back in a few hours)

---

@DilumAluthge do you happen to have a way to test if it actually works when a dependency has to be bumped? Some kind of example package? That way the secrets could be tested properly

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -7,6 +7,7 @@ on:
     secrets:
       DOCUMENTER_KEY:
         description: 'An SSH deploy key. See docs for more info.'
+        required: true
 jobs:
   CompatHelper:
     runs-on: ubuntu-latest
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   workflow_call:
     secrets:
-      COMPATHELPER_PRIV:
+      DOCUMENTER_KEY:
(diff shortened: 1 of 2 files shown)


.github/workflows/CompatHelper.yml
```diff
@@ -3,6 +3,11 @@ on:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
+  workflow_call:
+    secrets:
+      DOCUMENTER_KEY:
+        description: 'An SSH deploy key. See docs for more info.'
+        required: true
```
SaschaMann, line 10: Is this actually required? I'm never sure.

If not:

```suggestion
        required: false
```

DilumAluthge:
Would it be better to first get https://github.com/JuliaRegistries/CompatHelper.jl/pull/379 merged?

SaschaMann:
Oops, I forgot about that PR. The composite action is more flexible and there's not much point in having both in this case.

DilumAluthge:
What do end-users tend to prefer using? Composite workflows or re-usable workflows?

SaschaMann:
I don't really know, they're both relatively new so I'm not sure what people prefer.

The main advantage of the composite action would be that people can run things before or after it, the workflow is every so slightly shorter (doesn't require setting an OS for example).

PR #411, opened 2022-04-18
Allow `strict_version` entries

ven-k:
- `strict_version` helps to restrict the version bumps to exact `"= major.minor.patch"` 
- This is useful when you depend on packages that don't always follow the semver and you want automated test+bump for each release

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge", "Brown Center for Biomedical Informatics", "contributors"]
-version = "3.1.0"
+version = "3.2.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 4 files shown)


ven-k:
`Waiting for status to be reported` - which status is being expected here?

DilumAluthge:
I might be misunderstanding this feature, but I don't think this makes sense to include in CompatHelper.

The goal of CompatHelper is to notify package authors about breaking releases of

ven-k:
I want the versions in `MyPkg` to be bumped but something like `PkgA = "= 0.3.0"` be bumped to `PkgA = "= 0.3.1"` or `PkgA = "= 0.4.0"` ... i. e retain the {equals sign MAJOR.Minor.patch} in Project.toml compat entry

Currently If I've `PkgA = "= 0.3.0"` & `bump_compat_containing_equality_specifier=true`, it bumps to `PkgA = "0.4"` & drops `=` sign.

DilumAluthge:
It's been a couple years since this PR was first opened. When it was first opened, I was opposed to this feature. Since then, however, I've been thinking about it some more, and now I'm in favor of adding this feature.

It does look like we'll have to rebase and fix a lot of merge conflicts, or recreate the branch from scratch.

PR #418, opened 2022-05-16
Exclude yanked versions by default (but allow users to include yanked versions by setting `include_yanked = true`)

DilumAluthge:


src/dependencies.jl
```diff
@@ -89,8 +89,12 @@ function get_latest_version!(deps, registries)
                 versions_toml_path = joinpath(
                     registry, packages[uuid]["path"], "Versions.toml"
                 )
-                versions = VersionNumber.(collect(keys(TOML.parsefile(versions_toml_path))))
-
+                versions_dict = TOML.parsefile(versions_toml_path)
+                exclude_yanked = true
```
mattBrzezinski, line 93: Make this an kwarg possibly? Is there ever a time you would not want this to be `true` though? I can't imagine.

src/dependencies.jl
```diff
@@ -76,7 +76,7 @@ function clone_all_registries(f::Function, registry_list::Vector{Pkg.RegistrySpe
     end
 end
 
-function get_latest_version!(deps, registries)
+function get_latest_version!(deps, registries; exclude_yanked::Bool = true)
```
mattBrzezinski, line 79: Think we need to populate this all the way down from `main()` if we want to use it.

DilumAluthge: Let's do #419 first, which should make it a little easier to pass options around.

mattBrzezinski: Makes sense to me! I'll review it shortly.

DilumAluthge:
@mattBrzezinski @fchorney Which do you prefer for the public API?

1. `include_yanked` with a default of `false`
2. `exclude_yanked` with a default of `true`

fchorney:
> 1. `include_yanked` with a default of `false`
> 2. `exclude_yanked` with a default of `true`

I guess 1 possibly makes more sense, but I honestly have a hard time deciding if one is better than the other.

DilumAluthge:
bors merge

DilumAluthge:
@mattBrzezinski @fchorney Similar to #419, I'm going to merge this now so that we can get 3.2.0 registered by 00:00 UTC, but feel free to continue reviewing post-merge.

PR #419, opened 2022-05-16
Use an `Options` struct to pass options around, instead of needing to pass kwargs around everywhere

DilumAluthge:
Instead of needing to pass kwargs around everywhere, let's put all the kwargs into a single `Options` struct, and then we can just pass a single `options::Options` around, instead of needing to pass all of the kwargs around.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -4,9 +4,8 @@ function get_project_deps(
     api::Forge,
     ci::CIService,
     repo::Union{GitHub.Repo,GitLab.Project};
-    subdir::AbstractString="",
-    include_jll::Bool=false,
-    master_branch::Union{DefaultBranch,AbstractString}=DefaultBranch(),
+    options::Options,
+    subdir::AbstractString,
 )
     mktempdir() do f
         url_with_auth = get_url_with_auth(api, ci, repo)
@@ -14,13 +13,13 @@ function get_project_deps(
         @mock git_clone(url_with_auth, local_path)
 
         @mock cd(local_path) do
-            master_branch = @mock git_get_master_branch(master_branch)
+            master_branch = @mock git_get_master_branch(options.master_branch)
             @mock git_checkout(master_branch)
         end
 
         # Get all the compat dependencies from the local Project.toml file
         project_file = @mock joinpath(local_path, subdir, "Project.toml")
-        deps = get_project_deps(project_file; include_jll=include_jll)
+        deps = get_project_deps(project_file; include_jll=options.include_jll)
 
         return deps
     end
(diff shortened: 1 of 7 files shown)


src/dependencies.jl
```diff
@@ -4,23 +4,22 @@ function get_project_deps(
     api::Forge,
     ci::CIService,
     repo::Union{GitHub.Repo,GitLab.Project};
-    subdir::AbstractString="",
-    include_jll::Bool=false,
-    master_branch::Union{DefaultBranch,AbstractString}=DefaultBranch(),
+    options::Options,
+    subdir::String,
```
mattBrzezinski, line 8:
```suggestion
    subdir::AbstractString,
```

src/main.jl
```diff
@@ -45,54 +39,45 @@ Main entry point for the package.
...
+)
+    options = Options(; kwargs...)
+    return _true_main(env, ci_cfg; options)
+
+end
+
+function _true_main(
+    env::AbstractDict,
+    ci_cfg::CIService;
+    options::Options,
```
mattBrzezinski, line 52:
```suggestion
    kwargs...,
)
    options = Options(; kwargs...)
```

src/main.jl
```diff
@@ -45,54 +39,45 @@ Main entry point for the package.
...
-            subdir=subdir,
-            include_jll=include_jll,
-            master_branch=master_branch,
+            options,
+            subdir,
         )
 
-        if use_existing_registries
+        if options.use_existing_registries
             get_existing_registries!(deps, depot)
```
mattBrzezinski, line 68:
```suggestion
            get_existing_registries!(deps, options.depot)
```

mattBrzezinski (commented):
Might need to pull this locally down to review, GH makes it a bit tricky to review

DilumAluthge:
bors try

DilumAluthge:
@mattBrzezinski @fchorney I'm going to merge this now, just because I want to get #418 in soon, but feel free to continue reviewing post-merge; we can apply any necessary changes in follow-up PRs.

DilumAluthge:
bors merge

fchorney (approved):
Overall I like the idea, passing around all those options was definitely annoying. 

I'm a little unsure why we're using a pipe `|>` for `options.subdirs |> only` instead of just using `only(options.subdirs)` is there something that the pipe gives us that using parens doesn't?

DilumAluthge:
> I'm a little unsure why we're using a pipe `|>` for `options.subdirs |> only` instead of just using `only(options.subdirs)` is there something that the pipe gives us that using parens doesn't?

I just thought I'd try something new 😂 

If you found it confusing, that's probably a sign that we should switch back to the `only(options.subdirs)` syntax.

Issue #437, opened 2022-08-25
CompatHelper on GHA: Started erroring with pkg_server_registry_url not defined


kellertuer:
A few days ago our CompatHelper started erroring with the message that

```
ERROR: LoadError: UndefVarError: pkg_server_registry_url not defined
```

see
https://github.com/JuliaManifolds/Manifolds.jl/runs/8007521423?check_suite_focus=true#step:3:9

DilumAluthge:
It looks like you're using CompatHelper v2. Can you upgrade to CompatHelper v3?

kellertuer:
Ah, interesting, I will try, but I fear that will not run on a PR? I would prefer not to push such a test directly to master.

mattBrzezinski:
> Ah, interesting, I will try, but I fear that will not run on a PR? I would prefer not to push such a test directly to master.

Actions should run updated workflows in a PR.

kellertuer:
Sure, but this is just a scheduled GHA that does not run on push/pull/PR.

kellertuer:
Ah I could manually trigger running the GHA from that branch and it runs fine. So upgrading to version 3 resolved the problem.

DilumAluthge:
See also: https://github.com/JuliaRegistries/General/pull/67188

PR #438, opened 2022-08-27
Pin Julia version

rikhuijzer:
CompatHelper v2 doesn't work on Julia 1.8. This causes all workflow runs that are pinned to CompatHelper v2 to fail since the runners are allowed to use any Julia version (see also #437). To solve this problem in the future, this PR proposes to explicitly install Julia v1.8.

@DilumAluthge

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -10,16 +10,11 @@ jobs:
   CompatHelper:
     runs-on: ubuntu-latest
     steps:
-      - name: Check if Julia is already available in the PATH
-        id: julia_in_path
-        run: which julia
-        continue-on-error: true
-      - name: Install Julia, but only if it is not already available in the PATH
+      - name: Install Julia
         uses: julia-actions/setup-julia@v1
         with:
-          version: '1'
+          version: '1.8'
           arch: ${{ runner.arch }}
-        if: steps.julia_in_path.outcome != 'success'
       - name: "Add the General registry via Git"
         run: |
           import Pkg


DilumAluthge:
We set `version = "3"` when `Pkg.add`ing CompatHelper; that should be sufficient to ensure that CompatHelper v3 is installed, right?

CompatHelper v2 is end-of-life, so we don't support it anymore.

DilumAluthge:
I think the combination of fixing the `julia` compat in the General registry (https://github.com/JuliaRegistries/General/pull/67188) plus the `version = "3"` line (line 34 in https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml) should be sufficient here, right?

https://github.com/JuliaRegistries/CompatHelper.jl/blob/d5dc1f4179ab17499619a2d920d07fc3bab353fb/.github/workflows/CompatHelper.yml#L30-L35

rikhuijzer:
Changing 2 to 3 would be sufficient for all current setups yes. However, at this moment all CompatHelper setups pinned to v2 are broken and have to be manually updated to either go to CompatHelper v3 or to go back to Julia v1.7. The PR to the general registry doesn't solve that, right? 

This PR would avoid the problem for the future if it happens that v3 of CompatHelper is not compatible with, say, Julia 1.9.

The best solution would probably to add a new action `julia-actions/compathelper` which handles the Julia version and the communication with CompatHelper. Then it would be much less likely that some mass update for `CompatHelper.yml` is required.

DilumAluthge:
> The best solution would probably to add a new action `julia-actions/compathelper` which handles the Julia version and the communication with CompatHelper. Then it would be much less likely that some mass update for `CompatHelper.yml`-files is required.

Yeah, I think that's going to be the best long-term solution. I've been working on it here: https://github.com/JuliaRegistries/compathelper-action

It still needs some more testing though.

rikhuijzer:
> > The best solution would probably to add a new action `julia-actions/compathelper` which handles the Julia version and the communication with CompatHelper. Then it would be much less likely that some mass update for `CompatHelper.yml`-files is required.
> 
> Yeah, I think that's going to be the best long-term solution. I've been working on it here: https://github.com/JuliaRegistries/compathelper-action
> 
> It still needs some more testing though.

Ah so I see that you're only one year ahead of me with having the idea 😆

I'll start using it and let you know if I run into trouble

DilumAluthge:
I don't think that https://github.com/JuliaRegistries/compathelper-action (in its current form) actually solves the issue you're bringing up here.

Right now, `compathelper-action` is a composite action. Composite actions can only have `run` steps - they can't call other actions (like the `setup-julia` action). So `compathelper-action` isn't actually able to call `setup-julia` to install the appropriate version of Julia.

I think we'd need to convert `compathelper-action` from a composite action to a reusable workflow, and then it would be able to `setup-julia`.

DilumAluthge:
@rikhuijzer See https://github.com/JuliaRegistries/compathelper-workflow/pull/1

DilumAluthge:
It looks like reusable workflows won't work on self-hosted runners.

So for people using GitHub-hosted runners (which is most people), they can use the reusable workflow (https://github.com/JuliaRegistries/compathelper-workflow).

But for those people that are using self-hosted runners, they'll have to use the composite action (https://github.com/JuliaRegistries/compathelper-action).

@rikhuijzer Would you be interested in testing out both options (the reusable workflow and the composite action)?

rikhuijzer:
> Right now, `compathelper-action` is a composite action. Composite actions can only have `run` steps - they can't call other actions (like the `setup-julia` action). So `compathelper-action` isn't actually able to call `setup-julia` to install the appropriate version of Julia.

Are you sure about this? `julia-actions/cache` uses `actions/cache` (https://github.com/julia-actions/cache/blob/main/action.yml).

> @rikhuijzer Would you be interested in testing out both options (the reusable workflow and the composite action)?

Okay. I'll do that. It will take a bit though because I'm on the road next week for work

rikhuijzer:
> @rikhuijzer Would you be interested in testing out both options (the reusable workflow and the composite action)?

@DilumAluthge I've tested https://github.com/JuliaRegistries/compathelper-action manually and opened 2 PRs (https://github.com/JuliaRegistries/compathelper-action/pull/24 and https://github.com/JuliaRegistries/compathelper-action/pull/25).

The tests worked out fine for this Action. Would be nice if we could add some tests in CI, but without it things should also be okay given the relative simplicity of the setup. I'd say let's publish it and slowly move repositories over.

PR #449, opened 2023-02-09
Use the `RegistryInstances` API in `get_latest_version!` and `get_existing_registries!`

andreasnoack:
to make `use_existing_registries` work on more recent versions of Julia.

We use `RegistryInstances` to avoid using Pkg internals.


DilumAluthge:
Instead of using Pkg internals, can you use RegistryInstances.jl?

DilumAluthge:
We'll also need a compat entry for RegistryInstances?

src/dependencies.jl
```diff
@@ -75,26 +75,16 @@ function clone_all_registries(f::Function, registry_list::Vector{Pkg.RegistrySpe
...
-        registry_toml_path = joinpath(registry, "Registry.toml")
-        registry_toml = TOML.parsefile(registry_toml_path)
-        packages = registry_toml["packages"]
+function get_latest_version!(deps::Set{DepInfo}, registries::Vector{RegistryInstances.RegistryInstance}; options::Options)
+    for registry_instance in registries
+        packages = registry_instance.pkgs
 
         for dep in deps
-            uuid = string(dep.package.uuid)
+            uuid = dep.package.uuid
```
DilumAluthge, line 83: Can you add a type assertion here, to make sure that it's what we expect?

src/CompatHelper.jl
```diff
@@ -9,6 +9,7 @@ using Pkg: Pkg
 using TimeZones: TimeZones
 using TOML: TOML
 using UUIDs: UUIDs, UUID
+using RegistryInstances
```
DilumAluthge, line 12:
```suggestion
using RegistryInstances: RegistryInstances
```

src/dependencies.jl
andreasnoack: I don't think that field can be anything but an `UUID` https://github.com/JuliaRegistries/CompatHelper.jl/blob/cb50eedfe076b3045fc3a1da424011035435fa0e/src/utilities/types.jl#L10

src/CompatHelper.jl
andreasnoack: I'm not familiar with this pattern. What does it do?

DilumAluthge: It's equivalent to `import Foo`.

The idea is that we do not want to bring all of the exported names into scope.

andreasnoack: Makes sense. I thought you had to use import for that and indeed it's misleading if you don't plan to extend anything. I'll change this and also the tests tomorrow.

andreasnoack:
Aren't https://github.com/JuliaRegistries/CompatHelper.jl/blob/cb50eedfe076b3045fc3a1da424011035435fa0e/test/dependencies.jl#L40-L48 pseudo tests in that they don't actually check if there are cloned registries? Isn't the check in the current form just have you can push string to a vector? I tried to change `clone_all_registries` to use `reachable_registries` instead building a vector of paths but `reachable_registries` returns an empty list because `git_clone` doesn't seem to actually clone any registries. At least that's how I read the failure in https://github.com/JuliaRegistries/CompatHelper.jl/actions/runs/4144273670/jobs/7167123638#step:17:923

DilumAluthge:
Broadly speaking, we have two sets of tests here: the unit tests and the integration tests. The unit tests frequently use mocking (via Mocking.jl). In contrast, the integration tests use no mocking, and thus are always using real resources.

In the unit test you've linked, it looks like we mock the "git clone" functionality. So in its current state, this test is only testing the interface of function, not the behavior itself. The interface will be tested in the integration tests, but it's not always possible to cover all code paths in the integration tests (and we require 100% line coverage), so mocked unit tests can help hit other code paths.

For this specific unit test, yeah, I think it's fine to switch it from being mocked to not-mocked. You'd first modify https://github.com/JuliaRegistries/CompatHelper.jl/blob/cb50eedfe076b3045fc3a1da424011035435fa0e/test/dependencies.jl#L39 - for this test, we likely need to stop applying one or both of those Mocking.jl patches. Then, for the test, I think you'll need to do something like:
1. Save the original value of `Base.DEPOT_PATH`, etc.
2. mktempdir()` a new temp depot
3. Use Pkg to install more than one registry in that temp depot
4. Run the desired tests with that temp depot
5. Restore the original values of `Base.DEPOT_PATH`, etc.

DilumAluthge:
As far as where to get the multiple registries:
1. General
2. Maybe https://github.com/JuliaRegistries/Test?

andreasnoack:
@DilumAluthge I finally managed to get tests passing here.

DilumAluthge:
bors try

andreasnoack:
Done

DilumAluthge (approved):

DilumAluthge:
bors merge

PR #453, opened 2023-09-21
Fix Documenter to 0.27

lgoettgens:
This project is not compatible with Documenter 1.0 yet. So we should pin it to 0.27 for now, so that building the docs still succeeds.


DilumAluthge (approved):

DilumAluthge:
bors merge

PR #458, opened 2023-10-19
Open PRs for compat entries to weak dependencies

### LINKED ISSUE #452: Feature request: open PRs to update compat entries for weak dependencies (`[weakdeps]`)
the compats for `weakdeps` isn't getting any update.

sethaxen:
Fixes #452


sethaxen:
@DilumAluthge I'm having a little trouble understanding the test suite. Could you direct me to where I would need to make changes to test that an outdated version in `[weakdeps]` triggers a PR?

DilumAluthge:
So, the first step is to create some new "templates", which basically just represent what the `Project.toml` file would look like on the master branch of some Julia package.

The templates live here: https://github.com/JuliaRegistries/CompatHelper.jl/tree/master/test/templates

Then, for the integration tests, you add these new "template packages" here: https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/test/integration_tests.jl

That handles the integration tests. You'll likely also need to add some unit tests for the specific functions modified in this PR.

sethaxen:
@DilumAluthge I think this is now ready for review

src/utilities/new_versions.jl
```diff
@@ -54,7 +55,7 @@ function pr_info(
     pr_title_prefix::String,
 )
     new_pr_title = m"""
-    $(pr_title_prefix)CompatHelper: add new compat entry for $(name) at version
+    $(pr_title_prefix)CompatHelper: add new compat entry for $(name) in [$(section)] at version
```
DilumAluthge, line 58: I'd rather not change the PR title for regular deps. Can we keep the old PR title for regular deps, and only include the "in section" text for weak deps?

sethaxen: Done!

DilumAluthge:
@sethaxen Can you resolve the merge conflicts?

DilumAluthge:
@mattBrzezinski or @ericphanson Would you be able to take a first pass at reviewing this sometime?

DilumAluthge:
Let's also see if the integration tests pass.

bors try

mattBrzezinski:
> @mattBrzezinski or @ericphanson Would you be able to take a first pass at reviewing this sometime?

I _might_ have some time this week, but I can't make any promises. I'm also offline for the next two weeks following 😬 .

sethaxen:
The recent failures are unrelated to this PR. They come from Aqua v.8, which now enforces that extras have a compat entry (ironically, the failure is because Aqua itself does not have a compat entry).

src/dependencies.jl
```diff
@@ -18,35 +18,39 @@ end
...
+                    compat_entry = DepInfo(package)
+                    dep_entry = convert(String, strip(get(compat, name, "")))
+
+                    if !isempty(dep_entry)
+                        compat_entry.version_spec = semver_spec(dep_entry)
+                        compat_entry.version_verbatim = dep_entry
+                    end
+
+                    push!(project_deps, compat_entry)
+                    get!(dep_section, compat_entry, section)
```
ericphanson, line 47: what if a package is in multiple sections? E.g. `[deps]` and `[weakdeps]` is a [documented compatibility strategy](https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension)

ericphanson: If it’s only the title it seems like not too big of a deal either way, just wanted to make sure semantically it was OK. Maybe a simple choice is to just use deps in the title when present by looping over weak deps then deps (or vice versa if you prefer)

sethaxen: Ah, that's currently how this works. `get!(dep_section, compat_entry, section)` ensures that `dep_section[compat_entry]` is only set once to a section for a given package, and if the package was found in the `deps` section, it's already set, so `weakdeps` will be ignored.

ericphanson (approved):
This looks good to me. It seems slightly unfortunate to produce even more PRs that CompatHelper already does (i.e. we still don't have https://github.com/JuliaRegistries/CompatHelper.jl/issues/126) but that is kind of an orthogonal point.

It would be nice also for CI to be passing, if that is an easy fix to do here. I believe Bors will refuse to merge otherwise.

src/dependencies.jl
DilumAluthge: If possible, I'd like the titles for regular deps to remain the same as the old titles, and for the new title format to only be used for weakdeps.

sethaxen: @DilumAluthge this has already been handled in 3e52390

ericphanson:
Bump Project.toml version for a feature release (minor bump)? Then I think we're good to go.

ericphanson:
BTW I wish we could do incremental rollouts of github actions, e.g. deploy to 1% of users for a day or two to make sure there's no huge issue. Would be great for wide-impact tooling like CompatHelper. I guess we could do something really silly in the action like install a different version of CompatHelper.jl based on `rand(1:100)`? Not sure it's worth it.

DilumAluthge:
> BTW I wish we could do incremental rollouts of github actions, e.g. deploy to 1% of users for a day or two to make sure there's no huge issue. Would be great for wide-impact tooling like CompatHelper. I guess we could do something really silly in the action like install a different version of CompatHelper.jl based on `rand(1:100)`? Not sure it's worth it.

Yeah it would be nice to have some way if staggering the rollout.

ericphanson:
I guess the main thing is we don't get control over the install, as we have users do it themselves: https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml. If we had our own action that handled the install, then we could implement more sophisticated strategies in it. For example, we could push a release that hardcodes the date, and scales the probability of using the new version by the number of days since that date, so e.g. after 10 days, 100% of the time it uses the new release. If there was a problem, we could then push a new release that reverts to the old version etc. Then next release we could use the time-based rollout again, etc.

ericphanson:
bors merge

ericphanson:
I guess bors doesn’t listen to me anymore. Maybe @DilumAluthge can do the honors here

DilumAluthge:
@sethaxen Can you rebase this on the latest master?

sethaxen:
@DilumAluthge done!

PR #477, opened 2023-12-01
CI: migrate from Bors to GitHub Merge Queue; also, refactor some code, and add some more unit tests

DilumAluthge:
It looks like the public Bors instance has been shut down, so we are forced to migrate to GitHub Merge Queue instead.


.github/workflows/ci_unit.yml
```diff
@@ -1,11 +1,12 @@
...
 on:
+  merge_group: # GitHub merge queue
   pull_request:
   push:
     branches:
       - master
-      - staging
-      - trying
+      # - staging # Bors
+      # - trying # Bors
```
mattBrzezinski, line 9:
```suggestion
```

.github/workflows/ci_integration.yml
```diff
@@ -6,11 +6,12 @@ name: CI (integration tests)
...
 
 on:
+  merge_group: # GitHub merge queue
   push:
     branches:
       - master
-      - staging
-      - trying
+      # - staging # Bors
+      # - trying # Bors
```
mattBrzezinski, line 14:
```suggestion
```

mattBrzezinski (approved):

ericphanson:
How do we configure the merge queue? Did you already set it up to trigger integration tests?

mattBrzezinski:
> How do we configure the merge queue? Did you already set it up to trigger integration tests?

Screenshot 2023-12-01 at 17-56-15 Settings · Branches · Branch protection rule · JuliaRegistries_CompatHelper jl


My phone is upstairs, so I can't see if you're an admin in this repo, but here's the above settings. From skimming the [README](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) it seems like you just need to configure it here and in the workflow set `on -> merge_group`?

DilumAluthge:
Hmmm. Why isn't the merge queue working?

DilumAluthge:
I think that Merge Queue is unhappy because the integration tests aren't running on the PR.

But we intentionally don't want to run the integration tests on the PR, because we want to avoid hitting rate limits.

DilumAluthge:
I have unchecked the "Only merge non-failing pull requests" option in the merge queue settings.

DilumAluthge:
I think we basically just need to make sure that none of the required status checks are stuck at pending (yellow) on a PR. The easiest way to do that is to just fail those checks on PRs, which is what I've done here.

Because I've unchecked the "Only merge non-failing pull requests" option, those PR failures should not prevent us from adding a PR to the merge queue, at which point those same status checks will pass on the `merge_group` job.

ericphanson:
Looking at https://github.com/orgs/community/discussions/46757#discussion-4836009 and https://theunixzoo.co.uk/blog/2023-11-16-migrating-to-gh-merge-queues.html I think we do need a workaround here. But I think we should pass the checks on PRs rather than failing them, and require all checks pass

DilumAluthge:
My concern is that if we mark the integration tests as passed (green) on PRs, then I might look at a PR and mistakenly think to myself that the integration tests have run and passed, which would not be correct. At least with the red X, it's clear that the integration tests did not run and pass.

Maybe I can mark the integration tests as "skipped" on PRs. I'll try that out.

ericphanson:
Yeah, that makes sense. My concern is that contributors will see the red X and spend time trying to debug when there's actually nothing for them to do. It could also give reviewers the sense that a PR isn't ready, when maybe in fact it is. "skipped" seems the most accurate if we can get that to work.

ericphanson:
> My concern is that if we mark the integration tests as passed (green) on PRs, then I might look at a PR and mistakenly think to myself that the integration tests have run and passed, which would not be correct

This seems less problematic to me, as you can't merge directly (as far as I understand), you can only merge through the merge queue. So even if you think the pr is 100% ready to go, it will still go through the queue and the tests will still run. So we won't merge stuff that doesn't pass the tests.

DilumAluthge:
> and require all checks pass

Unfortunately, even if I'm able to get the integration tests to show up as "skipped", we still will have the Codecov status checks as red, because apparently the code coverage of the unit tests is only 99%, and we need the integration tests to get to 100%.

So the Codecov statuses are going to show up as red on PRs, unless we can figure out how to write some unit tests (maybe with the use of mocking) to get that last 1% of coverage.

ericphanson:
we can also do this: https://docs.codecov.com/docs/common-recipe-list#ease-target-coverage

DilumAluthge:
I want to keep enforcing that master stays at 100% code coverage.

ericphanson:
I think with this codecov setup we delete the `minimum` check

codecov.yml
```diff
@@ -1,7 +1,12 @@
...
+        branches:
+          - main
+          - gh-readonly-queue/main/*
+        target: 100%
+      # on PRs, we allow some leniency since integration tests haven't run yet
       default:
-        target: auto
-      minimum:
-        target: "100%"
+        target: 98%
```
DilumAluthge, line 12: @ericphanson This can be 99, right, not 98?

ericphanson: I dunno, I thought 98 was safer. We can try 99 though, if this merges down then it seems ok

ericphanson:
```suggestion
        target: 99%
```

codecov.yml
```diff
@@ -1,7 +1,12 @@
 coverage:
   status:
     project:
+      # no leniency on `main`
+      main:
+        branches:
+          - main
+          - gh-readonly-queue/main/*
```
DilumAluthge, line 8: @ericphanson Add a comment explaining that this is for GitHub Merge Queue?

codecov.yml
```diff
@@ -1,7 +1,12 @@
 coverage:
   status:
     project:
+      # no leniency on `main`
```
DilumAluthge, line 4:
```suggestion
      # no leniency on `main` (or in GitHub Merge Queue jobs)
```

codecov.yml
ericphanson:
```suggestion
          # for GitHub merge queue
          - gh-readonly-queue/main/*
```

ericphanson:
Screenshot 2023-12-02 at 19 05 31
can you remove this check @DilumAluthge ? then we can try this out

ericphanson:
ok, let's try it!

codecov.yml
```diff
@@ -1,7 +1,13 @@
 coverage:
   status:
     project:
+      # no leniency on `main` (or in GitHub Merge Queue jobs)
+      main:
+        branches:
+          - main
+          # for GitHub merge queue
+          - gh-readonly-queue/main/*
```
ericphanson, line 9:
```suggestion
      master:
        branches:
          - master
          # for GitHub merge queue
          - gh-readonly-queue/master/*
```

ericphanson:
Just realized trunk branch here is called `master` not `main`. Fixed

ericphanson:
ah, annoying - integration tests are not running here, bc they are running on the cancelled merge queue run from a few mins ago, and we have a global concurrency lock. I wonder if we can gate the lock after the `if` somehow so we only lock for actual runs

DilumAluthge:
The issue (I think) is that when you withdrew this PR from the merge queue, the integration tests job (on the merge_group event) did not get auto-cancelled. It's still running: https://github.com/JuliaRegistries/CompatHelper.jl/actions/runs/7071936138

We need to figure out how to auto-cancel merge_group jobs when a PR is withdrawn from the merge queue.

DilumAluthge:
Oh I see what you are saying: if the `if:` is going to evaluate to false anyway (on the PR build), why does it need to take the lock.

ericphanson:
> We need to figure out how to auto-cancel merge_group jobs when a PR is withdrawn from the merge queue.

For this it might be enough to turn on

>   cancel-in-progress: false

which we have in the integration tests. But I figured maybe we need that to not have dangling PRs and such on the integration repo.

ericphanson:
> Oh I see what you are saying: if the `if:` is going to evaluate to false anyway (on the PR build), why does it need to take the lock.

Actually, I think this might not matter too much, because even if the integration tests "ran" with this no-op `true` status, that would just get it into the queue where it still needs to wait for the old run to finish anyway, since in the queue we would do a "real" run. So probably there's little actual gains to be had here. The only actual speedup is if we can cancel in-progress runs then.

ericphanson:
Argh! It says the codecov target is 99% even in the merge queue:
Screenshot 2023-12-02 at 19 57 36

So my codecov.yml changes did not work as intended.

ericphanson:
I will remove from queue so we can try alternatives.

DilumAluthge:
Sometimes there's a delay before the Codecov status updates.

ericphanson:
I thought maybe it was ORing the two possibilities (99% or 100%) so I used a regex to try to make them more exclusive. The regex could be made more robust, but this should be enough to test the theory

ericphanson:
what if we just trigger integration tests some other way? e.g. adding a label to the PR. So they skip themselves (and fail) whenever the label is not present. Then when we're ready to run them, we add the label.

This is a security issue if we are restricting access to integration tests for security reasons (bc one could push bad ocmmits after the label is added). I don't think that is the case here where the integration tests do not have access to anything very sensitive. But maybe we could also have the label removed automatically when new commits are pushed if needed.

In this paradigm we wouldn't need merge queue at all, but we could still use it if we wanted.

DilumAluthge:
I sure do miss Bors.

The label idea won't be sufficient to get integration tests to run on PRs from forks.

ericphanson:
following the links there leads to https://github.com/imjohnbo/ok-to-test as well

DilumAluthge:
Codecov is the only problem here, right?

I think we should just get unit test coverage to 100%, and then use GitHub Merge Queue.

Or, alternatively, self-host our own instance of Bors.

DilumAluthge:
Alright, I have it down to only 2 uncovered lines (that are uncovered by the unit tests but covered by the integration tests). But I'm not sure how to write a unit test to cover those two lines, even with mocking.

I think, for the short-term, let's just revert to the original `codecov.yml`, and allow PRs to enter the merge queue even if the Codecov statuses are ❌.

We can maybe create a "developer docs" markdown page in the repo, and in that page, document that it's okay if Codecov shows up as 99% (and thus ❌) on a PR, because it will become 100% once the PR is added to the merge queue and integration tests run.

DilumAluthge:
> let's just revert to the original `codecov.yml`

#482

DilumAluthge:
> We can maybe create a "developer docs" markdown page in the repo, and in that page, document that it's okay if Codecov shows up as 99% (and thus ❌) on a PR, because it will become 100% once the PR is added to the merge queue and integration tests run.

#483

DilumAluthge:
Why is this not merging? I unchecked the "only merge passing PRs" option in the branch protection settings.

ericphanson:
are codecov checks required? maybe for required checks they must all be passing, or something like that

DilumAluthge:
That's really annoying if true. What, then, is the point of being able to uncheck the "only merge passing PRs" checkbox?

DilumAluthge:
I'm going to throw my hands up in defeat for now, and just lower the threshold to 99% (which is what the unit tests hit).

DilumAluthge:
Alright, the threshold for `codecov/project/minimum` is now 99%.

PR #480, opened 2023-12-02
add codecov threshold

ericphanson:
to help with https://github.com/JuliaRegistries/CompatHelper.jl/pull/477

docs: https://docs.codecov.com/docs/common-recipe-list#set-status-checks-to-block-coverage


DilumAluthge (changes requested):
Personally, I want us to keep enforcing that master must have 100% code coverage.

I think the better approach would be to figure out how to write unit tests to cover the last few lines that are uncovered by unit tests currently (but are covered by our integration tests).

ericphanson:
I wonder if this configuration would work for that?

codecov.yml
```diff
@@ -1,7 +1,16 @@
 coverage:
   status:
     project:
+      # no leniency on `main`
+      main:
+        branches:
+          - main
```
ericphanson, line 7: maybe this should be the merge-queue branch instead? but they don't support regex here :/

DilumAluthge:
Hmmm, so in the current state of this PR, I would allow the `merge_group` job to have code coverage less than 100%, right?

ericphanson:
I don't really understand if this is about the branch we are merging from (/ running on), or the branch we are merging into. Either way I think this is incomplete though. If it is the branch we are running on, then we want the merge-queue branch to have the strict checks, not just main itself. If it is the branch we are merging into, then this won't help for PRs to main.

codecov.yml
ericphanson:
```suggestion
        branches:
          - main
          - gh-readonly-queue/main/*
```

ericphanson:
Ok, I had to fix the syntax a bit, but this now validates:
```sh
curl -X POST --data-binary @codecov.yml https://codecov.io/validate
Valid!

{
  "coverage": {
    "status": {
      "project": {
        "main": {
          "branches": [
            "^main$",
            "^gh-readonly-queue/main/.*"
          ],
          "target": 100.0
        },
        "default": {
          "target": 98.0
        }
      }
    }
  }
}
```

DilumAluthge:
How would we go about testing this to make sure it works the way we want?

ericphanson:
I think the easiest way is just by merging things. We can test the happy path by merging this and your PR down. Then we can test the failure path by deleting the integration tests and seeing if that merges down. If so we can revert back.

PR #498, opened 2024-06-27
GitHub: get API url and server URL from env

gbruer15:
I encountered difficulties trying to use CompatHelper.jl on an Enterprise server, but it turns out I just needed to set the API url and the hostname correctly.

This pull request updates `auto_detect_ci_service` to automatically handle non-GitHub.com GitHub servers.

I am not sure how to add tests for this.

@DilumAluthge I see you are probably the most active here. Can you take a look at this and advise on tests?

### DIFF
--- a/src/utilities/ci.jl
+++ b/src/utilities/ci.jl
@@ -62,7 +62,10 @@ end
 
 function auto_detect_ci_service(; env::AbstractDict=ENV)
     if haskey(env, "GITHUB_REPOSITORY")
-        return GitHubActions()
+        api_hostname = get(env, "GITHUB_API_URL", "https://api.github.com")
+        clone_hostname = get(env, "GITHUB_SERVER_URL", "github.com")
+        clone_hostname = replace(clone_hostname, r"^(https?://)?" => "")
+        return GitHubActions(; api_hostname, clone_hostname)
     elseif haskey(env, "GITLAB_CI")
         return GitLabCI()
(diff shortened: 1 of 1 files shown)


DilumAluthge:
So, if we add this functionality into CompatHelper itself, it'll be a little difficult for us to test it.

Instead, could you modify your `CompatHelper.yml` file?

For example, if you look at the [recommended `CompatHelper.yml` file](https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml), part of it looks like this:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/97e9dcdde383ea5bea89d51de221698958b47630/.github/workflows/CompatHelper.yml#L37-L41

You could instead modify your `CompatHelper.yml` file to look like this:

```yaml
      - name: "Run CompatHelper"
        run: |
          import CompatHelper
          my_ci_cfg = CompatHelper.GitHubActions(;
              username = "...",
              email = "...",
              api_hostname = "...",
              clone_hostname = "...",
          )
          CompatHelper.main(ENV, my_ci_cfg)
        shell: julia --color=yes {0}
```

Where you'd fill in the relevant values in the `GitHubActions(; ...)` constructor.

That way, you can specify the exact values you need, without us needed to modify the source code of CompatHelper.jl (and thus needing to figure out a way to test it).

PR #502, opened 2024-09-05
Consider the extras block when checking for new versions

### LINKED ISSUE #166: check "extras" not just "deps" ?
would it make sense to check everything that's in "compat", not just the subset that's also listed in "deps"?  the stuff in "extras" seems not to be checked.  see https://github.com/GiovineItalia/Compose.jl/issues/374

andreasnoack:
With the work in https://github.com/JuliaRegistries/CompatHelper.jl/pull/458, it seems like it's pretty simple to close #166 as also conjectured in that issue.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -21,7 +21,7 @@ function get_project_deps(project_file::AbstractString; include_jll::Bool=false)
     dep_section = Dict{DepInfo,String}()
     project = TOML.parsefile(project_file)
 
-    for section in ["deps", "weakdeps"]
+    for section in ["deps", "weakdeps", "extras"]
         if haskey(project, section)
             deps = project[section]
             add_compat_section!(project)
--- a/test/dependencies.jl
+++ b/test/dependencies.jl
@@ -20,6 +20,8 @@ end
     for (k, s) in pairs(dep_section)
         if k.package.name ∈ ["Bex_jll", "Skix"]
             @test s == "weakdeps"
+        elseif k.package.name ∈ ["Baz"]
+            @test s == "extras"
         else
             @test s == "deps"
         end
@@ -31,6 +33,8 @@ end
     for (k, s) in pairs(dep_section)
         if k.package.name == "Skix"
             @test s == "weakdeps"
+        elseif k.package.name == "Baz"
+            @test s == "extras"
         else
             @test s == "deps"
         end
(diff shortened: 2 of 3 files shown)


DilumAluthge:
Should we make this toggleable? And should it be opt-in or opt-out?

DilumAluthge:
Some thoughts:
1. The main use case for `[extras]` is test-only deps. Most packages in the ecosystem currently don't have `[compat]` entries for their test-only deps, so this would lead to a surge of new PRs across the ecosystem.
2. AutoMerge requires packages to have `[compat]` entries for all regular deps. However, AutoMerge, doesn't currently require packages to have `[compat]` entries for test-only deps[^1]. CompatHelper was originally invented for the purpose of helping users meet the AutoMerge requirements. In some sense, that continues to be a core purpose of CompatHelper, so perhaps it makes sense that for the default behavior to be congruent with what AutoMerge requires (with of course the ability for users to opt-in to additional functionalities).

[^1]: Maybe we should add that requirement, but that's likely a discussion for a different day.

andreasnoack:
There is indeed a significant difference between direct package dependencies and test dependencies in that the latter can't break the environment alone, so having wrong compat (e.g. no) bounds on test dependencies could be considered a private matter. However, I don't really see why it was ever a good idea not to have compat bounds on test dependencies. It's not at all obvious that new versions of test dependencies wouldn't break your tests. However, bottomline here is probably that it should be opt-in. This will require a little more work on the PR, though.

andreasnoack:
@DilumAluthge has anything changed with respect to

> AutoMerge, doesn't currently require packages to have [compat] entries for test-only deps

or is it status quo?

DilumAluthge:
Nothing has changed, but we might consider making a breaking release of CompatHelper in which we start enforcing compat bounds for test deps (as the default behavior; we could add a kwarg that allows users to opt-out)?

nalimilan:
IMO it's important to file PRs to adjust bounds for extra deps by default. I just bumped into this case in StatsModels (https://github.com/JuliaStats/StatsModels.jl/pull/323) with CategoricalArrays 1.0. If CompatHelper doesn't notice developers that a new version of the dependency has been released, the package CI will most likely continue testing an old version forever, while most users will use the new one.

To minimize the number of new PRs, would it be possible to only track new versions when version bounds are specified for a given package?

devmotion:
I agree, I also just ran into this with CategoricalArrays 1.0 - tests were still running on an older version and CompatHelper didn't notify me about the release.

@DilumAluthge what's the blocker for the PR? That it would be a breaking change? Maybe it could be softened by a) only opening CompatHelper PRs for dependencies for which already a version bound exists (as suggested in the last comment) and/or b) making it an opt-in feature initially.

DilumAluthge:
> Maybe it could be softened by a) only opening CompatHelper PRs for dependencies for which already a version bound exists (as suggested in the last comment) and/or b) making it an opt-in feature initially.

I think either of those would be a good idea.

PR #505, opened 2024-09-23
Bugfix: in the second push (`force_ci_trigger()`), make sure to actually push to the SSH remote

### LINKED ISSUE #500: The "second push" (`force_ci_trigger()`) incorrectly pushes to the HTTPS remote, and as a result, Github Actions CI is not started for CompatHelper PRs
We are using `CompatHelper.jl` for our flow solver project [Trixi.jl](https://github.com/trixi-framework/Trixi.jl). However, since some time, our CI workflow does not start anymore for PRs opened by github-actions bot triggered by CompatHelper. 

We are using the workflow file linked on https://github.com/JuliaRegistries/CompatHelper.jl. We also use Documenter.jl and have `DOCUMENTER_KEY` set up as suggested. Our current workaround is to just close and reopen the PR, which then triggers CI. We also tried with a new Documenter key, which however did not help.

Do have any idea what could be in the way?

penelopeysm:
CompatHelper checks for the existence of an SSH key to determine an appropriate git remote URL:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/src/utilities/new_versions.jl#L254-L260

However, the `repo_git_url` variable generated here is not used later on. This means that if an SSH key is provided (via `COMPATHELPER_PRIV`) it is not actually used, because the remote URL is set as https here

https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/src/dependencies.jl#L6-L9

and never updated.

This PR fixes this by creating a new remote (by default called `compathelper-remote`) and calling `git remote add` before performing the rest of the git commands. If the remote already exists, we call `git remote set-url`. (This doesn't happen in a real run, but it does happen in the test suite.)

I do feel like ideally we'd use the ssh key and the url to clone the repo at the very start, but that requires reworking a fair bit more plumbing and I didn't want to shake it up too much. (For what it's worth, Documenter.jl uses the same approach as in the PR, i.e. adding a remote with the right URL.)

Since the signature of `force_ci_trigger` is changed, I also bumped the major version.

Fixes #500; I've tested this with my own fork and can confirm that this version of CompatHelper will generate PRs that trigger CI runs.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge", "Brown Center for Biomedical Informatics", "contributors"]
-version = "3.11.0"
+version = "4.0.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge", "Brown Center for Biomedical Informatics", "contributors"]
-version = "4.0.0"
+version = "3.12.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/src/utilities/git.jl
+++ b/src/utilities/git.jl
@@ -90,13 +90,13 @@ function git_get_master_branch(master_branch::DefaultBranch)
 end
 git_get_master_branch(master_branch::AbstractString) = master_branch
 
+function remote_exists(remote_name::AbstractString)
+    return remote_name in split(strip(read(`git remote`, String)), '\n')
+end
+
 function git_remote_add_or_seturl(remote_name::AbstractString, url::AbstractString)
-    remotes = split(strip(read(`git remote`, String)), '\n')
-    if remote_name in remotes
-        run(`git remote set-url $remote_name $url`)
-    else
-        run(`git remote add $remote_name $url`)
-    end
+    git_remote_subcommand = remote_exists(remote_name) ? "set-url" : "add"
+    run(`git remote $git_remote_subcommand $remote_name $url`)
 
     return nothing
 end
--- a/src/utilities/git.jl
+++ b/src/utilities/git.jl
@@ -89,3 +89,13 @@ function git_get_master_branch(master_branch::DefaultBranch)
     return string(strip(read(`git rev-parse --abbrev-ref HEAD`, String)))
 end
 git_get_master_branch(master_branch::AbstractString) = master_branch
+
+function git_remote_add_and_seturl(remote_name::AbstractString, url::AbstractString)
+    try
+        run(`git remote add $remote_name $url`)
+    catch
+        run(`git remote set-url $remote_name $url`)
+    end
+
+    return nothing
+end
--- a/src/utilities/git.jl
+++ b/src/utilities/git.jl
@@ -90,11 +90,12 @@ function git_get_master_branch(master_branch::DefaultBranch)
 end
 git_get_master_branch(master_branch::AbstractString) = master_branch
 
-function git_remote_add_and_seturl(remote_name::AbstractString, url::AbstractString)
-    try
-        run(`git remote add $remote_name $url`)
-    catch
+function git_remote_add_or_seturl(remote_name::AbstractString, url::AbstractString)
+    remotes = split(strip(read(`git remote`, String)), '\n')
+    if remote_name in remotes
         run(`git remote set-url $remote_name $url`)
+    else
+        run(`git remote add $remote_name $url`)
     end
 
     return nothing
--- a/src/utilities/new_versions.jl
+++ b/src/utilities/new_versions.jl
@@ -216,6 +216,7 @@ function make_pr_for_new_version(
     subdir::String,
     local_clone_path::AbstractString,
     dep_section::String,
+    remote_name::AbstractString="compathelper-remote",
 )
     if !continue_with_pr(dep, options.bump_compat_containing_equality_specifier)
         return nothing
@@ -266,6 +267,10 @@ function make_pr_for_new_version(
         master_branch_name = git_get_master_branch(options.master_branch)
         git_checkout(master_branch_name)
 
+        # Create a new remote (or update the given remote) with the appropriate
+        # URL (ssh/https)
+        git_remote_add_and_seturl(remote_name, repo_git_url)
+
         # Create compathelper branch and check it out
         new_branch_name = "compathelper/new_version/$(get_random_string())"
         git_branch(new_branch_name; checkout=true)
@@ -287,7 +292,7 @@ function make_pr_for_new_version(
             @info("Commit was a success")
             api_retry() do
                 @mock git_push(
-                    "origin", new_branch_name, pkey_filename; force=true, env=env
+                    remote_name, new_branch_name, pkey_filename; force=true, env=env
                 )
             end
 
@@ -297,7 +302,7 @@ function make_pr_for_new_version(
 
             options.cc_user && cc_mention_user(forge, repo, new_pr; env=env)
             options.unsub_from_prs && unsub_from_pr(forge, new_pr)
-            force_ci_trigger(forge, new_pr_title, new_branch_name, pkey_filename; env=env)
+            force_ci_trigger(forge, new_pr_title, remote_name, new_branch_name, pkey_filename; env=env)
 
             # Return to the master branch
             git_checkout(master_branch_name)
@@ -312,6 +317,7 @@ end
 function force_ci_trigger(
     api::GitLab.GitLabAPI,
     pr_title::AbstractString,
+    remote_name::AbstractString,
     branch_name::AbstractString,
     pkey_filename::Union{AbstractString,Nothing};
     env=ENV,
@@ -323,6 +329,7 @@ end
 function force_ci_trigger(
     api::GitHub.GitHubAPI,
     pr_title::AbstractString,
+    remote_name::AbstractString,
     branch_name::AbstractString,
     pkey_filename::Union{AbstractString,Nothing};
     env=ENV,
@@ -345,7 +352,7 @@ function force_ci_trigger(
         # Force push the changes to trigger the PR
         api_retry() do
             @debug "force_ci_trigger: force-pushing the changes to trigger CI on the PR"
-            @mock git_push("origin", branch_name, pkey_filename; force=true, env=env)
+            @mock git_push(remote_name, branch_name, pkey_filename; force=true, env=env)
         end
     end
 
--- a/src/utilities/new_versions.jl
+++ b/src/utilities/new_versions.jl
@@ -291,8 +291,10 @@ function make_pr_for_new_version(
         if commit_was_success
             @info("Commit was a success")
             api_retry() do
+                # For the first push, we intentionally push to the HTTPS remote (`origin`),
+                # because we want to avoid triggering duplicate CI runs.
                 @mock git_push(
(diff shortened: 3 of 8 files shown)


DilumAluthge:
If the user doesn't provide an SSH key, then CompatHelper pushes one time (using the HTTPS remote, and using the `GITHUB_TOKEN` to authenticate).

If the user does provide an SSH key, then CompatHelper actually does two pushes. This is intentional. Basically, it goes like this:
1. Push for the first time. It doesn't actually matter which remote we use this time.
2. Use the GitHub API to open a new PR (authenticating to the GitHub API using the `GITHUB_TOKEN`).
3. Amend the commit. (Just so that we can push again in the next step.)
4. Push again. This time, we push using the SSH remote, and using the SSH deploy key to authenticate.

This whole setup is intentional.

The basic problem is that we have to use the GitHub API to create the new PR (step 2 above). There is no other way to create the new PR, except the GitHub API. And the only way to authenticate to the GitHub API is to use the `GITHUB_TOKEN`. Recall that the `GITHUB_TOKEN` is automatically provided for us by GitHub Actions. Whenever we create a PR using the `GITHUB_TOKEN`, GitHub will not run CI jobs on that PR. This is an intentional limitation imposed by GitHub, and there is currently no way to change that behavior. Therefore, regardless of how we do the first push (in step 1 above), when we open the PR in step 2, CI will not be run on the PR, due to the limitation imposed by GitHub.

Therefore, it doesn't really matter what mechanism (HTTPS vs SSH) we use when we do the first push (in step 1 above), because no matter what, we are going to need to do the second push (in step 4 above). And it is that second push that uses the SSH remote; and because we are pushing to the SSH remote (and because we are authenticating using the SSH deploy key), that second push is able to trigger CI on our PR.

Does that make sense?

Basically, the first push (step 1 above) is done here:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/src/utilities/new_versions.jl#L289-L291

And the second push (step 4 above) is done here:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/src/utilities/new_versions.jl#L300-L300

Which then goes to this function:

https://github.com/JuliaRegistries/CompatHelper.jl/blob/e72bd6b815a69d14ccdb0c85b3589bed2a2af704/src/utilities/new_versions.jl#L323-L353

It's entirely possible that I'm misunderstanding your PR, and that my explanation above is not relevant to your PR - let me know if that's the case.

penelopeysm:
I understand the need for the second push! However, the second push doesn't actually use SSH. Although it provides a `GIT_SSH_COMMAND`, the remote URL is still a https one so `GIT_SSH_COMMAND` is never actually invoked.

I verified this by e.g. running `git remote -v` inside the `force_ci_trigger`. By default, GitHub Actions will censor the output because it contains the `GITHUB_TOKEN`, but you can either capture the output and check that it contains `https`, or (unsafe! :)) force it to print by e.g. reversing it or intercalating spaces.

Some debugging has gone on in #500 too – e.g. the ssh debug output is not printed even when setting `JULIA_COMPATHELPER_ENABLE_SSH_VERBOSE: true` indicating that `ssh -vvv` isn't actually being run, and separately I've tried setting `GIT_SSH_COMMAND` to `echo`, which should cause it to break if it's actually using the command, but it doesn't.)

DilumAluthge:
Ah, I see what you're getting at. The second push isn't actually using the SSH remote.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "CompatHelper"
 uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
 authors = ["Dilum Aluthge", "Brown Center for Biomedical Informatics", "contributors"]
-version = "3.11.0"
+version = "4.0.0"
```
DilumAluthge, line 4: I think this is a bugfix, right? I don't think we need to bump the major version. This was always supposed to be how CompatHelper works.

src/utilities/git.jl
```diff
@@ -89,3 +89,13 @@ function git_get_master_branch(master_branch::DefaultBranch)
     return string(strip(read(`git rev-parse --abbrev-ref HEAD`, String)))
 end
 git_get_master_branch(master_branch::AbstractString) = master_branch
+
+function git_remote_add_and_seturl(remote_name::AbstractString, url::AbstractString)
+    try
```
DilumAluthge, line 94: Instead of a try-catch, can we:

1. First, get the list of all existing remotes.
2. Then, check if `remote_name' is in the list of existing remotes.
3. Then, decide which command to run based on the result from step 2.

src/utilities/new_versions.jl
```diff
@@ -216,6 +216,7 @@ function make_pr_for_new_version(
     subdir::String,
     local_clone_path::AbstractString,
     dep_section::String,
+    remote_name::AbstractString="compathelper-remote",
```
DilumAluthge, line 219: Does this need to be a configurable parameter? I think it's fine for us to just set `CONST COMPATHELPER_REMOTE = "compathelper-remote"` and then use that.

Project.toml
penelopeysm: Hmm I guess it depends on whether you consider the individual function signatures to form part of the public API (=> major), or if it's just the overall effect of running the main function (=> bugfix). Happy to go either way on this 🙂

src/utilities/new_versions.jl
```diff
@@ -287,7 +292,7 @@ function make_pr_for_new_version(
             @info("Commit was a success")
             api_retry() do
                 @mock git_push(
-                    "origin", new_branch_name, pkey_filename; force=true, env=env
+                    remote_name, new_branch_name, pkey_filename; force=true, env=env
```
DilumAluthge, line 295: We actually want this "first push" to go to the HTTPS remote (not the SSH remote). That way, we can avoid triggering duplicate CI runs.

src/utilities/new_versions.jl
```diff
@@ -312,6 +317,7 @@ end
 function force_ci_trigger(
```
DilumAluthge, line 317:
```suggestion
const COMPATHELPER_SSH_REMOTE_NAME = "compathelper-ssh-remote"

function force_ci_trigger(
```

src/utilities/new_versions.jl
```diff
@@ -345,7 +352,7 @@ function force_ci_trigger(
         # Force push the changes to trigger the PR
         api_retry() do
             @debug "force_ci_trigger: force-pushing the changes to trigger CI on the PR"
-            @mock git_push("origin", branch_name, pkey_filename; force=true, env=env)
+            @mock git_push(remote_name, branch_name, pkey_filename; force=true, env=env)
```
DilumAluthge, line 355:
```suggestion
            @mock git_push(COMPATHELPER_SSH_REMOTE_NAME, branch_name, pkey_filename; force=true, env=env)
```

src/utilities/new_versions.jl
```diff
@@ -323,6 +329,7 @@ end
 function force_ci_trigger(
     api::GitHub.GitHubAPI,
     pr_title::AbstractString,
+    remote_name::AbstractString,
```
DilumAluthge, line 332: I don't think we need to accept this as an argument.

Project.toml
DilumAluthge: I also think we can get away without modifying any function signatures, at least not of public functions.

src/utilities/new_versions.jl
DilumAluthge, line 295:
```suggestion
                    "origin", new_branch_name, pkey_filename; force=true, env=env
```

src/utilities/new_versions.jl
```diff
@@ -287,7 +292,7 @@ function make_pr_for_new_version(
             @info("Commit was a success")
             api_retry() do
                 @mock git_push(
```
DilumAluthge, line 294:
```suggestion
                # For the first push, we intentionally push to the HTTPS remote (`origin`),
                # because we want to avoid triggering duplicate CI runs.
                @mock git_push(
```

Project.toml
penelopeysm: Gotcha!

src/utilities/new_versions.jl
```diff
@@ -297,7 +302,7 @@ function make_pr_for_new_version(
 
             options.cc_user && cc_mention_user(forge, repo, new_pr; env=env)
             options.unsub_from_prs && unsub_from_pr(forge, new_pr)
-            force_ci_trigger(forge, new_pr_title, new_branch_name, pkey_filename; env=env)
+            force_ci_trigger(forge, new_pr_title, remote_name, new_branch_name, pkey_filename; env=env)
```
DilumAluthge, line 307:
```suggestion
            force_ci_trigger(forge, new_pr_title, new_branch_name, pkey_filename; env=env)
```

DilumAluthge:
@penelopeysm Thanks for finding this bug, by the way!

test/patches.jl
```diff
@@ -23,6 +23,7 @@ git_push_patch = @patch function CompatHelper.git_push(
 end
 git_gmb_patch = @patch CompatHelper.git_get_master_branch(master_branch) = nothing
 git_checkout_patch = @patch CompatHelper.git_checkout(branch) = nothing
+git_remote_add_and_seturl_patch = @patch CompatHelper.git_remote_add_and_seturl(remote_name, url) = nothing
```
DilumAluthge: What is the value of `remote_name` here?

src/utilities/git.jl
```diff
@@ -90,11 +90,12 @@ function git_get_master_branch(master_branch::DefaultBranch)
 end
 git_get_master_branch(master_branch::AbstractString) = master_branch
 
-function git_remote_add_and_seturl(remote_name::AbstractString, url::AbstractString)
-    try
-        run(`git remote add $remote_name $url`)
-    catch
+function git_remote_add_or_seturl(remote_name::AbstractString, url::AbstractString)
+    remotes = split(strip(read(`git remote`, String)), '\n')
```
DilumAluthge, line 94: To make testing a little easier, can you factor this out into a separate function named `remote_exists(remote_name::AbstractString)` that returns a Bool?

test/patches.jl
penelopeysm: It'd be `compathelper-ssh-remote`, though actually I realise this isn't needed at all, it's not being used anywhere.

penelopeysm:
I think I've addressed everything, and added some small unit tests :)

Given that we've added new exposed functions here (the ones in `utilities/git.jl`, should this be a minor patch release?

test/utilities/git.jl
```diff
@@ -362,3 +362,43 @@ end
...
+    mktempdir() do f
+        cd(f) do
+            run(`git init`)
+            run(`git remote add origin foo-url`)
+
+            @testset "set existing remote" begin
+                url = "foo-url"
+                CompatHelper.git_remote_add_or_seturl("origin", "foo-url")
+                output = strip(read(`git remote get-url origin`, String))
+                @test output == url
```
DilumAluthge, line 394: Can you now do `git_remote_add_or_seturl()` with a different URL value, and then confirm that the new URL was set correctly?

penelopeysm: Oops, that's silly of me.

DilumAluthge: Sorry, I should clarify. You first have a test that sets the URL to `bar-url` and checks that it was set correctly. Now, immediately following that test, can you add a second test that changes the URL to `bar-url2` and checks that the change went through correctly?

penelopeysm: Can do!

DilumAluthge (approved):

DilumAluthge:
Thank you @penelopeysm!

penelopeysm:
No problem, it's a very useful tool :)

Thank you for taking the time to go through this!

DilumAluthge (approved):

DilumAluthge:
https://github.com/JuliaRegistries/General/pull/115803

PR #518, opened 2024-12-15
Add comments about deploy keys

ericphanson:
Clarifies their purpose, why there is `DOCUMENTER_KEY` and `COMPATHELPER_PRIV`, and link to docs

This workflow is linked to from the README/docs as an example so these comments should help users.

xref https://julialang.slack.com/archives/C6A044SQH/p1734169106738789

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -41,5 +41,10 @@ jobs:
         shell: julia --color=yes {0}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # This repo uses Documenter, so we can reuse our [Documenter SSH key](https://documenter.juliadocs.org/stable/man/hosting/walkthrough/).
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):

ericphanson:
I think this is uncontroversial so I'll add it to the merge queue, but if there is post-merge feedback I'll gladly make a new PR to address it

DilumAluthge:
I was going to ask if some version of this content should be added to the README. But as you point out, we already link to this YAML file from the README, so it may be sufficient to have the comments in the YAML file.

@ericphanson Should we also update PkgTemplates.jl so that the templated `CompatHelper.yml` file that PkgTemplates creates (in new packages) contains these same comments?

ericphanson:
that would be good, but they probably can't be the exact same since whether or not you have a documenter key depends on how you configure the template. So I think the comments would either need to be slightly more generic / not always applicable, or templated.

PR #525, opened 2025-05-12
Update GitHub action workflows

abhro:
- Use julia-actions/cache instead of actions/cache for less boilerplate
- Use julia-actions/docdeploy instead of manually specifying julia commands
- Use `shell:` option for cleaner job step specifications
- Update codecov/codecov-action to use newest version (v5)
  - Update options for codecov-action


abhro:
Gentle bump @DilumAluthge

PR #528, opened 2025-08-13
Add options for handling `[extras]` dependencies

### LINKED ISSUE #166: check "extras" not just "deps" ?

devmotion:
This PR extends #502 based on @DilumAluthge's suggestions in https://github.com/JuliaRegistries/CompatHelper.jl/pull/502#issuecomment-3183867774.

I added an option `open_prs_for_extras` that can be set to `IfExistingCompatExtras()` (the default), `AllExtras()` or `NoExtras()` [my gut feeling is that the names could be improved a bit]. With `IfExistingCompatExtras()`, only `[extras]` dependencies with existing compat entry are updated; with `AllExtras()`, all `[extras]` dependencies are updated (also those without existing compat entry); and with `NoExtras()`, `[extras]` dependencies are not updated.

I went with singleton types (basically a type-based enum) because 1) it felt more consistent with the existing `EntryType` and 2) it's not necessary to check whether the provided value is valid (which would be necessary if the API would be based on `Symbol`s).

Fixes #166. Closes #502.

### DIFF
--- a/test/main.jl
+++ b/test/main.jl
@@ -62,7 +62,7 @@
                     "GITHUB_REPOSITORY" => "CompatHelper.jl", "GITHUB_TOKEN" => "token"
                 ) do
                     prs = CompatHelper.main()
-                    @test length(prs) == 3
+                    @test length(prs) == 4
                     @test prs[1] isa GitHub.PullRequest
                     @test prs[2] isa GitHub.PullRequest
                 end
@@ -97,7 +97,7 @@
                     ) do
                         delete!(ENV, "GITHUB_REPOSITORY")
                         prs = CompatHelper.main()
-                        @test length(prs) == 3
+                        @test length(prs) == 4
                         @test prs[1] isa GitLab.MergeRequest
                         @test prs[2] isa GitLab.MergeRequest
                     end


devmotion:
I added a note.

src/main.jl
```diff
@@ -37,6 +38,13 @@ Main entry point for the package.
 - `cc_user=false`: CC the user on the pull requests
 - `bump_version=false`: When set to true, the version in Project.toml will be bumped if a pull request is made. Minor bump if >= 1.0, or patch bump if < 1.0
 - `include_yanked=false`: When set to true, yanked versions will be included when calculating what the latest version of a package is
+- `open_prs_for_extras::ExtrasType=IfExistingCompatExtras()`: How to handle dependencies in `[extras]`.
+  Can be `IfExistingCompatExtras()` (only open a pull request for a dependency in `[extras]` if a compat entry for it exists),
+  `AllExtras()` (open pull requests for all dependencies in `[extras]`),
+  or `NoExtras()` (do not open pull requests for dependencies in `[extras]`).
+
+!!! note
+    The `open_prs_for_extras` keyword argument is considered experimental (not public API) and may be changed or removed in a non-breaking release.
```
DilumAluthge, line 47:
```suggestion
    The `open_prs_for_extras` keyword argument is considered experimental (not public API) and may be changed or removed in a non-breaking release. Same goes for `AllExtras`, `IfExistingCompatExtras`, `NoExtras`, and `ExtrasType`.
```

test/deps/Project.toml
```diff
@@ -14,6 +13,10 @@ Skix = "3db6da90-6ed3-11ee-0779-f549c8e3e90d"
 [extensions]
 Ext = ["Bex_jll", "Skix"]
 
+[extras]
+Baz = "ea10d353-3f73-51f8-a26c-33c1cb351aa5"
```
DilumAluthge, line 17: Any particular reason for moving Baz from `[deps]` to `[extras]`?

src/CompatHelper.jl
```diff
@@ -12,6 +12,7 @@ using UUIDs: UUIDs, UUID
 using RegistryInstances: RegistryInstances
 
 export DropEntry, KeepEntry, NewEntry
+export AllExtras, IfExistingCompatExtras, NoExtras
```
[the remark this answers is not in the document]
devmotion:
```suggestion
```

src/main.jl
```diff
@@ -15,6 +15,7 @@
         cc_user=false,
         bump_version=false,
         include_yanked=false,
+        open_prs_for_extras::ExtrasType=IfExistingCompatExtras(),
```
devmotion, line 18:
```suggestion
        open_prs_for_extras::CompatHelper.ExtrasType=CompatHelper.IfExistingCompatExtras(),
```

DilumAluthge:
I can try to review this weekend.

test/deps/Project.toml
DilumAluthge: I'm just coming back to this now. I realize now that if we remove `Baz` from `[deps]`, then `[deps]` doesn't contain any direct non-stdlib non-JLL dependencies (because `Foobar_jll` is a JLL, and `LinearAlgebra` is a stdlib), which obviously is a codepath that we want to make sure to cover.

Can you move Baz back to `[deps]`, and then have an additional dep in `[extras]`?

DilumAluthge:
> I can try to review this weekend.

Famous last words, huh?

devmotion:
@DilumAluthge can you create a new release?

PR #530, opened 2025-09-20
docs: added instructions to avoid 403 error when using CompatHelper with Github Workflows

lfenzo:
See related discussions in:

- https://discourse.julialang.org/t/developing-package-compathelper-the-requested-url-returned-error-403/99881
- https://discourse.julialang.org/t/compathelper/115090

I also ran into this issue when using GitHub Workflows. Adding a note in the docs could save others some time.

cc.: @DilumAluthge


JohannesMorkrid (approved):
Thanks alot! I used 3-4 hours trying to fix this (after abandoning fixing it 2 months ago). Please consider if there is another way to make CompatHelper.yml easier to use as it was default in the PkgTemplates when I started my code structure.

DilumAluthge (approved):

